PR #22, opened 2014-12-11
Don't throw an error if the symlink already exists

vikhyat:
I'm running Unicorn in development with two worker processes, which is causing the symlink creation to be attempted twice.

### DIFF
--- a/lib/ember-cli/app.rb
+++ b/lib/ember-cli/app.rb
@@ -43,7 +43,10 @@ module EmberCLI
     end
 
     def symlink_to_assets_root
-      assets_path.join(name).make_symlink dist_path.join("assets")
+      begin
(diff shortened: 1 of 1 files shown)


vikhyat:
Still getting this error after your change to remove the rescue, and even when I bump the number of unicorn workers down to 1. Probably a race condition between the unicorn master and forked worker.

If you don't want to use a `rescue` maybe wrap it in a mutex?

rwz:
Well, crap. If it still blows up given that if condition, mutex won't help. It definitely needs a rescue.

rwz:
Thanks for reporting that. Again. :)

rwz:
Should be fixed in 0.0.15

vikhyat:
Thank you! :smile:

PR #27, opened 2014-12-19
Compile ember app only upon request.

rondale-sc:


rwz:
I'm gonna close this for now. We can return to this later to try another approach.

PR #43, opened 2015-01-04
Update README.md

bradurani:
Updated README.md to reflect a complication that tripped up my own usage of this gem.

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,11 +115,10 @@ Your Ember application will now be served at the `/` route.
 
 ## Additional Information
 
-EmberCLI Rails runs `ember build` with the `--output-path` and `--watch` flags
-on. The `--watch` flag sets tells EmberCLI to watch for file system events and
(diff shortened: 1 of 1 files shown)


rondale-sc:
Why remove the small snippet about the `--watch` flag?

bradurani:
I see, whether it uses run or compile is indeed dependent on `Rails.env.development?` but that's in the middleware. If `consider_all_requests_local = false` the middleware is never used, see [here](https://github.com/rwz/ember-cli-rails/blob/master/lib/ember-cli-rails.rb#L31) and [here](https://github.com/rwz/ember-cli-rails/blob/master/lib/ember-cli/helpers.rb#L25). As a result, the only time Ember CLI is built is in `rake assets:precompile`, where the `--watch` flag is not used. If you think it's worth it I can revise the PR to reflect this complexity but I don't want to confuse anyone.

rondale-sc:
No, that should be fine.  Thanks for the quick followup.  :)

LGTM

/cc @rwz

PR #45, opened 2015-01-04
Update README.md Additional Information

bradurani:
Update README.md to mention that `config.consider_all_requests_local = true` must be set in `development.rb` for Ember CLI to automatically build.


rwz:
Just FYI you don't have to close and re-open pull-requests every time you need to make a change. You can just push to the branch or even rewrite commits and github will update pull-request accordingly automatically.

PR #47, opened 2015-01-06
fix showing the required ember-cli npm version

jhsu:
with the incorrect version installed, an undefined error is thrown:

```
NameError (undefined local variable or method `requirement' for #<EmberCLI::App:0x007feaf89b0df8>):
  ember-cli-rails (0.0.14) lib/ember-cli/app.rb:111:in `check_ember_cli_version!'
  ...
```


rwz:
Man, we really need to write some test for that :(

Thanks for the fix!

rwz:
Released 0.0.15

rondale-sc:
@rwz Yeah, totally.  Some simple tests might be good, but I'm not sure how we'd setup a test harnes that'd cover all the versions of ember-cli and Rails that we support.  :(  If you have any ideas, I'd really like some coverage.

PR #50, opened 2015-01-06
use local ember executable as fallback

fivetanley:
uses the local ember executable in node_modules/.bin
in one of the apps if it is available. This is useful
for production environments where you cannot install
packages globally into the PATH with npm install -g.

### DIFF
--- a/lib/ember-cli/app.rb
+++ b/lib/ember-cli/app.rb
@@ -68,9 +68,19 @@ module EmberCLI
       MSG
     end
 
+    def ember_path
+      @ember_path ||= begin
+        local_path = app_path.join('node_modules', '.bin', 'ember')
+        if File.exist?(local_path) && File.executable?(local_path)
+          local_path
+        else
+          fail "No local ember executable found. You should run `npm install` inside the #{@app} app located at #{app_path}"
+        end
+      end
+    end
+
     private
 
-    delegate :ember_path, to: :configuration
     delegate :match_version?, :non_production?, to: Helpers
     delegate :tee_path, to: :configuration
     delegate :configuration, to: :EmberCLI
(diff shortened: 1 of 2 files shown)


rwjblue:
We should be using the local version of ember for each individual app (and not using the global at all).

fivetanley:
Doesn't the local version work the same way as the global version and just tap into the local code? Like grunt/grunt-cli?

rwjblue:
@fivetanley - Yeah, until we change it in some random version and everything breaks ;)

fivetanley:
@rwjblue updated

fivetanley:
I guess i need to look through and see what's still using the config `ember_path`.

fivetanley:
Looks like it was only used in app, so I went ahead and removed it.

fivetanley:
I also updated this so that jruby support worked in rails 4.1. JRuby does not support options to `system`. You'll get a warning for `error: :out` but not for chdir, and assets will continue to compile.

rondale-sc:
@fivetanley Neat!  You are usin' this with jRuby?  Has it been alright?

fivetanley:
Other than the chdir hiccup it seems to be working ok!

rwjblue:
:+1: from me...

lib/ember-cli/app.rb
```diff
@@ -15,7 +15,9 @@ def initialize(name, options={})
     def compile
       prepare
       silence_stream STDOUT do
-        system(env_hash, command, chdir: app_path, err: :out)
+        Dir.chdir app_path do
```
rwz: Neat, I didn't know about `Dir.chdir`!

rwz:
Released 0.0.16! :tada:

PR #52, opened 2015-01-07
Add LiveReload Instructions to README

iheanyi:
I figured others were probably wondering how to get LiveReload enabled for this app, so I just went ahead and added a quick summary on how to do so. Hope this is helpful to you all!

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,6 +113,19 @@ could render your app at the `/` route with the following view:
 
 Your Ember application will now be served at the `/` route.
 
+## Enabling LiveReload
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -113,6 +113,19 @@ could render your app at the `/` route with the following view:
...
+
+```
+guard 'livereload' do
+  ...
+  watch(%r{(app/{<your-appname>/\w+/(.+\.(js|hbs|html)))})
+  ...
+end
+```
+
+This tells Guard to watch your EmberCLI app directory for any Javascript, Handlebars, or HTML changes and reloading the browser on any file changes.
```
rondale-sc: Might want to mention that the file extensions will need to be modified if they are using different extensions (for example if you add ember-cli-coffee you'd want .coffee files to trigger a change)

iheanyi: Valid point, I'll add that as well. I also realized that I forgot some CSS/SCSS file extensions as well.

rwz:
Thanks!

Issue #51, opened 2015-01-07
Ember-cli-rails compiles wrong asset files


andruby:
I've spent hours figuring out why my "rake assets:precompile" suddenly started failing :)

[/lib/ember-cli/app.rb:142](https://github.com/rwz/ember-cli-rails/blob/b698902ac293340d408a7a5d404686adcbf5efb7/lib/ember-cli/app.rb#L142) seems to be the root cause: 
```ruby
Rails.configuration.assets.precompile << /(?:\/|\A)#{name}\//
```

This adds **all folders with the same name as the app** to the precompile list. I happened to have a css folder with the same name as an ember-cli app. This caused a compilation error because the css files in that folder depend on sass variables defined elsewhere (which are normally imported upfront when compiling the usual application.css file).

Can we change the line in question so that it only adds the specific ember-cli asset folders?

andruby:
I've alleviated my specific problem with [this commit](https://github.com/andruby/ember-cli-rails/commit/fe4e0aef0154ff38da5bcbbe113c14b368f822af). It adds :appname/ to the precompile list, not */:appname/. With that commit users can create asset subfolders with the :appname, but not asset "root" folders with the :appname.

Can we explicitly add files to the precompile list instead of directories?  So far I've noticed that these files are precompiled:
* :app/vendor.(js|css|map)
* :app/:app.(js|css|map)
* :app/failed.png
* :app/passed.png
* :app/test-loader.js
* :app/test-support.(css|js|map)

rwz:
Well, short answer is no. We can't make an explicit list of files to precompile, because EmberCLI plugins can actually implicitly add files to build output directory and we don't have a clear way to track that.

I don't really remember why I implemented it this particular way, but for now I don't see a problem with just changing precompile regex to `/\A<your-app-name>\//`.

Do you see any problems with that?

The assumption is that when you have an EmberCLI app called "frontend" for instance, everything in `/assets/frontend` is going to be compiled and you as a developer are going to try to avoid using this namespace for anything that's not related to this particular app.

andruby:
I've created a pull request that changes the regex to /\A<your-app-name>\// at #53.

This does fix our case, where we use two rather generic ember app names (`gate` and `venture`) and we define the css for the individual items in assets/modules/<venture>/. 

Thanks!

rwz:
Alright, seems like this one is solved. See #53.

PR #53, opened 2015-01-08
Add :appname/ to the precompile list instead of */:appname/. refs #51.

andruby:
As discussed with @rwz in #51, this commit no longer causes subfolders with the same name as an app to be compiled.


rwz:
Released 0.0.17

PR #54, opened 2015-01-08
update documentation around jquery-rails

fivetanley:
This was a good couple of hours trying to chase this down between ember-cli issues and this repository. Seems like it should be documented and also not the default case for Ember CLI apps.

Related https://github.com/rwz/ember-cli-rails/issues/32

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,11 +23,15 @@ EmberCLI addons too! Without further ado, let's get in there!
 
 ## Installation
 
-Firstly, you'll have to include the gem in your `Gemfile` and `bundle install`
+Firstly, you'll have to include the gem in your `Gemfile`
 
 ```ruby
 gem "ember-cli-rails"
 ```
+
+Next, remove `jquery-rails` from your `Gemfile`. If you want to keep jquery-rails
+around for other JavaScript bundles, Ember CLI will detect jquery-rails in your
+gemfile and exclude it from the Ember apps' builds if you leave it in the `Gemfile`.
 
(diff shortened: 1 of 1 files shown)


rondale-sc:
@fivetanley Definitely.  This is something we want to make explicit.  Or at least make a way to disable this behavior if necessary.   The perhaps incorrect assumption was that user would include jquery with their rails assets and we'd save space, but we've found several cases where this isn't desired.

rondale-sc:
@fivetanley I'm not sure that we should advocate that the user remove `jquery-rails` as it is likely used elsewhere.  I'm going to try to add a flag to the initializer tonight to disable the check for jquery-rails so that ember will include jquery when the flag is set.   Hopefully that'll help

fivetanley:
I think the behavior of not including jquery is very confusing, especially as it is the default in ember-cli and also not included in any docs/warnings in the rails console. I feel like excluding jquery from ember should be opt-out and in the documentation.

rwjblue:
Im not really sure what you mean. If jquery-rails gem is present, we use that jquery and tell ember-cli to not concat into its vendor.js. Otherwise we would be including jQuery twice.

rondale-sc:
@rwjblue The scenario that I'm worried about is

  - User has jquery-rails and uses it in parts of their app (or as we recently found out has it as a dependency of another gem)
  - User includes their ember app onto a page that doesn't also include applicaton.js

In this scenario since the Jquery constant is found we tell ember not to send it which breaks their app (unexpectedly).

PR #56, opened 2015-01-09
Broccoli fingerprints

chrisdevor:
For those looking to use the broccoli/ember fingerprinting and asset pipeline, I've created the following branch that may be helpful to others.  The basic idea here is that after ember compiles, we copy everything over to the rails assets folder.  

When done in production mode, it works as follows:
1. Compile and fingerprint rails assets w/o any ember assets
2. Compile and fingerprint ember assets outside of rails
3. Copy the compiled ember assets into the rails asset directory (public/assets)
4. Update rail's manifest.json to reference ember's app_name.js, app_name.css, vendor.js and vendor.css.

This might be helpful for this issue:  https://github.com/rwz/ember-cli-rails/issues/30

I'd be happy to hear other people's thoughts.

Side note: I also disabled the `SUPPRESS_JQUERY` flag, which seemed to help our app.

### DIFF
--- a/lib/ember-cli/app.rb
+++ b/lib/ember-cli/app.rb
@@ -76,6 +76,19 @@ module EmberCLI
       end
     end
 
+    def dist_path
+      @dist_path ||= EmberCLI.root.join("apps", name).tap(&:mkpath)
+    end
+
+    def assets_path
+      @assets_path ||= EmberCLI.root.join("assets").tap(&:mkpath)
+    end
+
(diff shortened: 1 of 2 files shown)


chrisdevor:
Setting `fingerprint: { [...] generateRailsManifest: true }` in our `brocfile.js` is a better way than this PR...  

These aren't the commits you're looking for...  You can go about your business... Move along...

PR #59, opened 2015-01-15
Expose a rake task to `npm install`

seanpdoyle:
Assumes either:
* `bower` is a globally accessible command
* `package.json` includes:
```json
"tasks": {
  "postinstall": "/path/to/bower/bin/bower install"
}
```

Corresponds to https://github.com/rwz/ember-cli-rails/issues/33

### DIFF
--- a/lib/ember-cli/app.rb
+++ b/lib/ember-cli/app.rb
@@ -17,7 +17,7 @@ module EmberCLI
     end
 
     def install_dependencies
-      exec "npm install"
+      exec "#{npm_path} install"
     end
 
     def run
@@ -82,7 +82,7 @@ module EmberCLI
     private
 
     delegate :match_version?, :non_production?, to: Helpers
-    delegate :tee_path, to: :configuration
+    delegate :tee_path, :npm_path, to: :configuration
     delegate :configuration, to: :EmberCLI
 
     def build_timeout
--- a/lib/ember-cli/app.rb
+++ b/lib/ember-cli/app.rb
@@ -16,6 +16,10 @@ module EmberCLI
       silence_stream(STDOUT){ exec command }
     end
 
+    def install_dependencies
+      exec "npm install"
+    end
+
     def run
       prepare
       cmd = command(watch: true)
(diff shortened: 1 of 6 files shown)


rondale-sc:
@seanpdoyle This looks really promising.  I'm going to test this today or tomorrow and talk to @rwz about it. 

/cc @rwz

seanpdoyle:
Great! I really dislike the fact that you need to depend on `postinstall` pointing to a bower, but its a start

lib/ember-cli/helpers.rb
```diff
@@ -25,10 +25,18 @@ def non_production?
       !Rails.env.production? && Rails.configuration.consider_all_requests_local
     end
 
+    def inject_install_dependencies_task!
```
rwz: Could we move the definition if this task into a separate file? Kind if like the way it's described [here](http://blog.nathanhumbert.com/2010/02/rails-3-loading-rake-tasks-from-gem.html)

lib/ember-cli/app.rb
```diff
@@ -16,6 +16,10 @@ def compile
       silence_stream(STDOUT){ exec command }
     end
 
+    def install_dependencies
+      exec "npm install"
```
rwz: How about we use our `which` helper to locate the `npm` bin here? We could cache it in configuration as well.

Something like this:

```ruby
class EmberCLI::Configuration
  def npm_path
    @npm_path ||= which("npm")
  end
end

class EmberCLI::App
  delegate :npm_path, to: :configuration
  def install_dependencies
    exec "#{npm_path} install"
  end
end

rwz:
Looks good to me!

rondale-sc:
:+1:

rwz:
Yay! Now I guess we need to describe this process in README somehow.

rwz:
Thanks forPR @seanpdoyle

rondale-sc:
@seanpdoyle Yeah, thanks!  :beers:

PR #62, opened 2015-01-20
Raise error when app timeout to compile

josemarluedke:
Why? It's really annoying to wait for the app to load and never load, so after a while I realize that I need to go to the terminal and check if the ember app have been compiled. Plus, if you have Travis CI or anything else running your tests and you have an error in your ember app, your tests might not get it because nothing raises.

### DIFF
--- a/lib/ember-cli/app.rb
+++ b/lib/ember-cli/app.rb
@@ -68,6 +68,8 @@ module EmberCLI
 
         ============================= WARNING! =============================
       MSG
+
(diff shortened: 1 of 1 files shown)


rondale-sc:
We are actively working on this.  See, https://github.com/rwz/ember-cli-rails/issues/23 for more details.  Definitely a pain point though.

Cheers

rwz:
Check out the latest 0.1.0 release. It might solve your problem in a slightly different way.

josemarluedke:
Awesome to see it implemented already. I will check that out. I guess you can close this PR if you want to.

PR #63, opened 2015-01-23
Use non_production? helper in middleware

decasia:
The `Helpers.non_production?` method correctly handles cases where local development is happening outside of a Rails environment called `development`. It seems to make sense to use it in the middleware code, no?

### DIFF
--- a/lib/ember-cli/middleware.rb
+++ b/lib/ember-cli/middleware.rb
@@ -15,7 +15,7 @@ module EmberCLI
 
     def enable_ember_cli
       @enabled ||= begin
-        if Rails.env.development?
+        if Helpers.non_production?
           EmberCLI.run!
         else
           EmberCLI.compile!


pedrocarrico:
@decasia Wouldn't this mean that in a `staging` environment we would be changing the current behaviour?

decasia:
@pedrocarrico, not necessarily. non_production? reads:

```
      !Rails.env.production? && Rails.configuration.consider_all_requests_local
```

So just don't set `consider_all_requests_local` and it will still use production behavior, right?

decasia:
If you are setting up a `staging` environment to be like prod, you wouldn't want that flag set anyway, I imagine.

rwz:
> It seems to make sense to use it in the middleware code, no?
Well, no. The middleware only kicks in if `non_production?` is true. And then, it needs to perform different things for development and testing. So, the assumption that `Helpers.non_production?` is equal to `!Rails.env.development?` is incorrect.

decasia:
OK, I'll open an issue to discuss other options.

PR #69, opened 2015-01-26
Change livereload path in README to <ember-cli-path>/app

jesenko:
Current watch expression includes `~/tmp` and `~/dist` paths. As files in those two folders also get modified after recompilation, additional unnecessary reloads are triggered.

Changing watch path to `<ember-cli-path>/app` still reloads for the majority of relevant changes; however some changes will not trigger reload (e.g. `Brocfile.js`, etc), but as I recall changes in these files also don't trigger reload when running standalone EmberCLI.

Alternatively, one could exclude `~/tmp` and `~/dist` paths explicitly, however I think it is not worth the trouble...

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,14 +123,15 @@ init` and then add the following to your `Guardfile`.
 ```ruby
 guard "livereload" do
   # ...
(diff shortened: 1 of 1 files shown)


rwz:
Thanks!

rondale-sc:
+1

PR #77, opened 2015-02-07
Improve RACK_ENV=production rake assets:precompile

seanpdoyle:
* Update the README
 * use project local `bower` install
 * mention removing `js_compressor = :uglifier`
* Simplify rake task declaration

### DIFF
--- a/README.md
+++ b/README.md
@@ -158,15 +158,21 @@ install`:
 gem "rails_12factor", group: :production
 ```
 
-Add a `package.json` to the root of your Rails project. This is to make sure
-it'll be detected by the NodeJS buildpack.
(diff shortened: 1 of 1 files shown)


rwz:
I don't feel comfortable suggesting to disable uglifier step. I was actually going to suggest removing minification on EmberCLI side and try to deal with slow deployments some other way (haven't have time to look into that yet, sadly). Let hang on this for a while and think of a better/other way of doing this. I'll keep this PR open in case we won't find anything. :)

The improvements to the Rake tasks are baller though. Could you maybe extract them into separate PR, so we could merge them right away?

seanpdoyle:
:+1:   #79

PR #79, opened 2015-02-09
Cleanup Rake Tasks

seanpdoyle:
* We don't need to require the sprockets tasks
* `override_assets_precompile_task` is an overcomplication of `task foo:
 :bar`
* We should make `ember-cli:compile` depend on
 `ember-cli:install_dependencies`
 * Worst case scenario, the `npm install` is a no-op


rwz:
:+1: :tada: Thanks!

rondale-sc:
@seanpdoyle Nice! :beers:

PR #84, opened 2015-02-13
Support Gemfile in ember-cli application

sevos:
Some EmberCLI addons require gems like compass or susy to be installed.
There is no need to have these gems in the rails project's Gemfile.
This pull request adds the support for Gemfile in the ember-cli
application's directory.

Tested with:

  - rvm 1.26.10
  - ruby 2.1.3
  - rubygems 2.4.5

Caveats:
  - the ruby version in both ember-cli and rails applications must be
  identical

### DIFF
--- a/lib/ember-cli/app.rb
+++ b/lib/ember-cli/app.rb
@@ -20,6 +20,7 @@ module EmberCLI
     end
 
     def install_dependencies
+      exec "bundle install" if gemfile_path.exist?
       exec "#{npm_path} install"
     end
 
@@ -241,7 +242,12 @@ module EmberCLI
       ENV.clone.tap do |vars|
         vars.store "DISABLE_FINGERPRINTING", "true"
         vars.store "EXCLUDE_EMBER_ASSETS", excluded_ember_deps
+        vars.store "BUNDLE_GEMFILE", gemfile_path
       end
+    end
+
+    def gemfile_path
+      app_path.join('Gemfile')
     end
 
     def exec(cmd, options={})


pedrocarrico:
:+1:

pedrocarrico:
@sevos Won't `exec` fail silently if there's no Gemfile in the `app_path` directory?
Perhaps you could make the inclusion of a Gemfile optional if there are no ruby dependencies in your ember-cli application.
Also, can I suggest a fixup/amend of your last commit?

sevos:
Please do not merge yet, it does not work on my workmate's machine with almost identical setup. Investigating ;)

sevos:
I will rebase and squash commits when ready

sevos:
@pedrocarrico - ok squashed and tested

rwz:
Just add these gems to your Rails app Gemfile with `require: false` option.

sevos:
@rwz this is not DRY solution, because these deps belong to the frontend project, not to the backend project (we keep these in two separate repositories).

rwz:
Alright, let's rethink this. Does this PR works ok on your machine?

sevos:
yup, and on all dev machines of guys from my team (so far, one is on holiday). Tested on zsh and bash (with RVM only, no rbenv unfortunately). 

Bundler uses the variable to select Gemfile to use. If the variable is not set it looks for the file in current directory (in that case is the rails' project directory, where you start the webserver) and sets the variable to point to the Gemfile in current directory, so subsequent bundler loads (in child processes for example) do not have to search for the Gemfile. In our case we need to override this variable when executing commands in ember-cli app context, so if it needs to load up some gems, it uses proper Gemfile.

lib/ember-cli/app.rb
```diff
@@ -241,9 +244,14 @@ def env_hash
       ENV.clone.tap do |vars|
         vars.store "DISABLE_FINGERPRINTING", "true"
         vars.store "EXCLUDE_EMBER_ASSETS", excluded_ember_deps
+        vars.store "BUNDLE_GEMFILE", gemfile_path
       end
     end
 
+    def gemfile_path
+      File.join(app_path, 'Gemfile')
```
rwz: Could you use Pathname API here instead?

```ruby
app_path.join("Gemfile")
```

lib/ember-cli/app.rb
```diff
@@ -20,6 +20,9 @@ def compile
     end
 
     def install_dependencies
+      if File.exists?(gemfile_path)
```
rwz: Given the gemfile is a Pathname object, this could be

```ruby
exec "bundle install " if gemfile_path.exist?
```

sevos:
done :)

rwz:
:+1:

sevos:
Sorry, I should have  tested this last change, check #86

seanpdoyle:
Hi @sevos.

This gem is approaching a `1.0.0` release, which gives us a great opportunity to make breaking changes.

We're considering removing  `Gemfile` and `bundle install` support for EmberCLI applications.

Would you mind pointing me to the addons you use that motivated this PR?

PR #86, opened 2015-02-13
Fix Pathname to String conversion

sevos:


rwz:
:((

That's dumb. Pathnames act as String pretty much everywhere, I wonder what's the big deal with this case. Anyways, thanks for the heads up.

sevos:
Just for the record:

```ruby
2.2.0 :006 >   project = Pathname.new(File.expand_path("~/projects/core"))
 => #<Pathname:/Users/sevos/projects/core>
2.2.0 :007 > env = {}
 => {}
2.2.0 :008 > env.store "VARIABLE", project
 => #<Pathname:/Users/sevos/projects/core>
2.2.0 :009 > system env, "echo $VARIABLE"
TypeError: no implicit conversion of Pathname into String
	from (irb):9:in `system'
	from (irb):9
	from /Users/sevos/.rvm/rubies/ruby-2.2.0/bin/irb:11:in `<main>'
```

sevos:
The problem is that it implements only `#to_s` method where probably it should also implement `#to_str` method which is responsible of implicit conversion of an object to string: http://stackoverflow.com/questions/11182052/to-s-vs-to-str-and-to-i-to-a-to-h-vs-to-int-to-ary-to-hash-in-ruby

PR #90, opened 2015-02-22
Serve EmberCLI tests as mountable engine

### LINKED ISSUE #18: Make original ember tests available through rails routes
Make http://localhost:4200/tests/* available through rails somehow.

//cc @mwoods79

seanpdoyle:
Closes #18

Expose EmberCLI test suite through Rails

### DIFF
--- a/lib/ember-cli-rails.rb
+++ b/lib/ember-cli-rails.rb
@@ -1,4 +1,4 @@
-require "ember-cli/railtie" if defined?(Rails)
+require "ember-cli/engine" if defined?(Rails)
 
 module EmberCLI
   extend self
@@ -15,6 +15,10 @@ module EmberCLI
 
   def configuration
     Configuration.instance
+  end
+
+  def get_app(name)
+    configuration.apps[name]
   end
 
   def prepare!
--- a/README.md
+++ b/README.md
@@ -113,6 +113,28 @@ could render your app at the `/` route with the following view:
 
 Your Ember application will now be served at the `/` route.
 
+## Ember Test Suite
+
+To run an Ember app's tests in a browser, mount the `EmberCLI::Engine`:
+
+```ruby
+# config/routes.rb
+
+Rails.application.routes.draw do
+  unless Rails.env.production?
+    mount EmberCLI::Rails => 'tests'
+  end
+
+  root 'application#index'
+end
+```
+
+Ember tests are served based on the route you mount the Engine on (in this
+example, `/tests`) and the name of the Ember app.
+
+For example, to view tests in `app/frontend`,
+visit `http://localhost:3000/tests/frontend`.
+
 ## Enabling LiveReload
 
 In order to get LiveReload up and running with EmberCLI Rails, you can install
--- a/app/controllers/ember_tests_controller.rb
+++ b/app/controllers/ember_tests_controller.rb
@@ -1,0 +1,27 @@
+class EmberTestsController < ActionController::Base
+  def index
+    render text: test_html_with_corrected_asset_urls, layout: false
+  end
+
+  private
+
+  def test_html_with_corrected_asset_urls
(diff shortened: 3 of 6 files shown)


seanpdoyle:
Unfortunately, this doesn't yet work.

The `app/frontend/tests/index.html` is being served properly from the `EmberCLI` dist_path, but the script and link tags are pointing to assets like `assets/vendor.css`, instead of `assets/frontend/vendor-FINGERPRINT.css`.

Any ideas? cc: @rwjblue @rwz @rondale-sc

rwz:
This is amazing. Thanks for this great work!

We could maybe perform a naive gsub before serving the html to replace `assets/` with `assets/:app_name/`. That's easy and solves the problem.

rwz:
Also, why remove `railties` from dependiencies? `Rails::Engine` is a part of `railties` as well AFAIK.

lib/ember-cli-rails.rb
```diff
@@ -1,4 +1,4 @@
-require "ember-cli/railtie" if defined?(Rails)
+require "ember-cli/engine" if defined?(Rails)
```
seanpdoyle: @rwz if the railties / engine is only loaded when `Rails` constant exists, we're really dependent on Rails, not railties.

If Rails is included in the dependencies (which it should be, this being `ember-cli-rails` and all), then engines and railties come along for the ride, right?

If not, I'll gladly add it back to the gemspec

rwz:
Also, what do you think about implicitly mounting the engine in development environment? But maybe use less obvious namespace, like `/ember-cli-tests/:app_name` instead? Then we could save some user's effort setting it up and not be a nuisance to those who don't care about this feature.

lib/ember-cli-rails.rb
rwz: `Rails` constant is defined in `railties`. So in this particular case we're depending on either `rails` or `railties`. I don't really have a hard preference on either of these, but one of them definitely should be there.

seanpdoyle:
I am against implicitly mounting the engine.

What happens if people want to configure the URL?

We could add another configuration value:

```ruby
EmberCLI.configure do |config|
  config.tests_root = "/my-ember-cli-tests"
end
```

Or we could just depend on the pre-existing API provided by Rails engines.

For users who don't care about this feature, they're opted out by default, and don't have to worry about our obscure URL convention conflicting with their existing routes.

Also, requiring an explicit mount frees us from worrying about catch-all routes.

```ruby
# config/routes.rb

# this would have precedent over our route
match "*route-to-ember" => "ember#index"

# this would match before the catchall
mount EmberCLI::Engine => "tests"
match "*route-to-ember" => "ember#index"
```

seanpdoyle:
It's worth noting that `<script src="testem.js">` is 404ing, since it isn't in the asset pipeline.

This doesn't seem to affect the test runnner, though

rwz:
It seems to be an empty dummy file: https://github.com/ember-cli/ember-cli/blob/v0.1.15/lib/broccoli/testem.js

Seems like we could safely ignore the fact it's missing. The other option would be to do something like this:

```ruby
get 'testem', to: -> { [200, {}, [""]] }, format: :js
```

rwz:
... or even better, handle it in the middleware. :)

seanpdoyle:
For the time being, I'm ok with it 404ing.

rwz:
Thanks again, you did an awesome job implementing this feature. :+1: :+1: :+1: 

Do you want to be a maintainer if the gem? You seem to care and put in a ton of work implementing features. We could definitely use all the help we could get from highly motivated people like you :)

seanpdoyle:
@rwz I'd love to!

rwz:
@seanpdoyle done.

rondale-sc:
@seanpdoyle @rwz :+1: :beers:

PR #91, opened 2015-02-24
Add app/**/* and config/**/* to gemspec

jesenko:
When installing latest release v0.1.7 from ruby gems, files from those two directories are missing, thus mounting Ember tests does not work.


rwz:
Oops! That's embarassing. I've tested everything locally with Gemfile pointing to a local dir and totally forgot about gemspec. Thanks for the fix!

rwz:
Released 0.1.8

PR #99, opened 2015-03-05
Add Support for Rails 3.1

seanpdoyle:
Add Support for Rails 3.1

* reduce Railties version
* only inflect on `CLI` acronym if that API is available

### DIFF
--- a/ember-cli-rails.gemspec
+++ b/ember-cli-rails.gemspec
@@ -12,6 +12,6 @@ Gem::Specification.new do |spec|
 
   spec.required_ruby_version = ">= 1.9.3"
 
-  spec.add_dependency "railties",        "~> 4.0"
-  spec.add_dependency "sprockets-rails", "~> 2.0"
+  spec.add_dependency "railties", ">= 3.1", "< 5"
+  spec.add_dependency "sprockets", ">= 2.0"
 end
--- a/lib/ember-cli/engine.rb
+++ b/lib/ember-cli/engine.rb
@@ -5,8 +5,10 @@ module EmberCLI
     end
 
     initializer "ember-cli-rails.inflector" do
-      ActiveSupport::Inflector.inflections :en do |inflect|
-        inflect.acronym "CLI"
+      if Rails.version > "3.2"
+        ActiveSupport::Inflector.inflections :en do |inflect|
+          inflect.acronym "CLI"
+        end
       end
     end
 


lib/ember-cli/app.rb
```diff
@@ -95,10 +95,14 @@ def ember_path
 
     private
 
-    delegate :match_version?, :non_production?, to: Helpers
+    delegate :match_version?, to: Helpers
     delegate :configuration, to: EmberCLI
     delegate :tee_path, :npm_path, :bundler_path, to: :configuration
 
+    def non_production?
+      !Rails.env.production?
```
rwz: Relying on `Rails.env.production?` is an anti-pattern and totally now gonna work for any production-like environment with a different name. For instance lots of people actually create a separate environment called `staging` by copying `production.rb` into `staging.rb` and deploing it. This change ensures that none of this gonna work anymore. This is the main reason `Helpers.non_production?` exists in the first place.

seanpdoyle: Could we expose these are configurations, with the default value set from the logic in `Helpers`? 

The project that is driving these changes has `config.consider_all_requests_local = false`

rwz: How do you mean?

`config.consider_all_requests_local = false` is usually a sign of a production-like environment. Which means that you probably want a production-like behavior from the gem (do nothing, let assets be precompiled).

seanpdoyle: Fair enough.

The problem I'm facing is that I've manually set `config.use_ember_middleware = true`, overriding the value of `non_production?`. This works for browsing the app, the EmberCLI app's `test/index.html` is no longer included in the build, since `non_production?? "development" : "production"` is evaluating to `"production"`.

This is breaking the work done in `0.1.7`

seanpdoyle:
@rwz I've responded in https://github.com/rwz/ember-cli-rails/pull/99#discussion_r25913281 but it's now hidden by an outdated diff

lib/ember-cli/app.rb
rwz: Errm... does it mean that you want ember tests in production?

rwz: I mean, you want ember tests engine to work in an environment that has `config.use_ember_middleware = true`?

seanpdoyle: @rwz yes.

I've toggled that setting to false in the project, but I'm not sure if it'll be accepted.

Either way, we might want to clarify how we determine the Ember ENV?

rwz: Ok, this is a bigger issue that it seems.

The gem has essentially three modes of operation:
- development
  Starts a build server on first request, runs it until the process dies.
- test
  Builds assets once, does nothing after.
- production
  Does nothing.

I guess instead of having this logic being spread out in lots of different places, we could make a single config option to let user set the operation mode manually.

rwz: Also, this discussion doesn't have anything to do with 3.1 support, it's a separate issue, am I wrong?

seanpdoyle: @rwz that's probably true.

cc: @rondale-sc 

In the mean time, what part of this diff is a point of contention to be being merged?

seanpdoyle: Should I reduce the scope of this PR to just the gemspec and inflector changes?

rwz: Yeah, that'd be cool. Let's remove the version changes, and precompile flag from it.

ember-cli-rails.gemspec
```diff
@@ -12,6 +12,6 @@ Gem::Specification.new do |spec|
 
   spec.required_ruby_version = ">= 1.9.3"
 
-  spec.add_dependency "railties",        "~> 4.0"
-  spec.add_dependency "sprockets-rails", "~> 2.0"
+  spec.add_dependency "railties",  ">= 3.1"
```
rwz: Let's do "> 3.1", "< 5", since we can't use "~>"

lib/ember-cli/engine.rb
```diff
@@ -4,9 +4,11 @@ class Engine < Rails::Engine
       ActionView::Base.send :include, ViewHelpers
     end
 
-    initializer "ember-cli-rails.inflector" do
-      ActiveSupport::Inflector.inflections :en do |inflect|
-        inflect.acronym "CLI"
+    if ActiveSupport::VERSION::MAJOR > 3 && ActiveSupport::VERSION::MINOR > 1
```
rwz: Could we do something less verbose here like `Rails.version > "3.2"`?

Not sure is `Rails.version` is a thing it 3.1.*, but there probably is something.

lib/ember-cli/engine.rb
```diff
@@ -4,9 +4,11 @@ class Engine < Rails::Engine
       ActionView::Base.send :include, ViewHelpers
     end
 
-    initializer "ember-cli-rails.inflector" do
-      ActiveSupport::Inflector.inflections :en do |inflect|
-        inflect.acronym "CLI"
+    if Rails.version >= "3.2"
```
seanpdoyle: @rwz :+1:

lib/ember-cli/engine.rb
```diff
@@ -6,7 +6,7 @@ class Engine < Rails::Engine
 
     initializer "ember-cli-rails.inflector" do
       ActiveSupport::Inflector.inflections :en do |inflect|
```
seanpdoyle: @rwz actually, `.inflections` is a problem as well, since i *doesn't* take arguments in 3.1

rwz:
So, can you please undo all the changes that are not gemspec and inflection? And then we can merge it.

rwz:
Oh, and change version comparison to `> "3.2"`.

seanpdoyle:
@rwz good to go

rwz:
:+1:

PR #100, opened 2015-03-05
Adds seanpdoyle as gem author

seanpdoyle:


rwz:
A post-merge question. Are you sure you want to use your thoughtbot email?

I assume that you're not gonna be a Thoughtbot employee forever, so why not use a personal email here instead? :)

PR #108, opened 2015-03-12
Add details on sending CSRF to README

thewatts:
I was receiving this error when trying to create a record on the server, had forgotten that Rails (by default) requires CSRF.

```
Can't verify CSRF token authenticity
Completed 422 Unprocessable Entity in 1ms

ActionController::InvalidAuthenticityToken
```

Added a little "how-to" to the README (mainly because the [`ember-rails`]( gem automatically sets up the configuration necessary for CSRF when initializing an app, so for a new person they may not realize they are missing this step.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,6 +113,35 @@ could render your app at the `/` route with the following view:
 
 Your Ember application will now be served at the `/` route.
 
+## CSRF Tokens
+
+Your Rails controllers, by default, are expecting a valid authenticity token to be submitted with non-`GET` requests.
+Without it you'll receive a `422 Unprocessable Entity` error, specifically: `ActionController::InvalidAuthenticityToken`.
+
+In order to add that token to your requests, you need to add into your template:
+
+```erb
+<!-- /app/views/application/index.html.erb -->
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -113,6 +113,35 @@ could render your app at the `/` route with the following view:
 
 Your Ember application will now be served at the `/` route.
 
+## CSRF Tokens
+
+Your Rails controllers, by default, are expecting a valid authenticity token.
```
seanpdoyle: It might be worth mentioning that CSRF is for non-`GET` requests.

thewatts: Good call there @seanpdoyle !

seanpdoyle:
thanks @thewatts!

LGTM

cc: @rwz @rondale-sc

rondale-sc:
LGTM

At some point (soon) we should probably talk about getting a proper documentation site ready.  Readme is getting pretty long.

thewatts:
@seanpdoyle && @rondale-sc most welcome :)

Thank you folks for doing all the groundwork!

PR #110, opened 2015-03-12
Remove rake_tasks block in engine

BlakeWilliams:
When the `rake_tasks` block is present and loading tasks it causes the
tasks to be loaded twice and runs them twice each time they are called.

Since ember-cli-rails uses an engine instead of a Railtie, it will load tasks for us: https://github.com/rails/rails/blob/4e0ec961e1f81ef672e2728b8650e4358962962f/railties/lib/rails/engine/configuration.rb#L52
https://github.com/rails/rails/blob/4e0ec961e1f81ef672e2728b8650e4358962962f/railties/lib/rails/engine.rb#L32


rwz:
:+1:

PR #115, opened 2015-03-18
skipping inflections for Rails 3.2.x versions too

2468ben:
I was running Rails version 3.2.18 and still getting the inflector error, so just wanted to skip it for 3.2.x

### DIFF
--- a/lib/ember-cli/engine.rb
+++ b/lib/ember-cli/engine.rb
@@ -5,7 +5,7 @@ module EmberCLI
     end
 
     initializer "ember-cli-rails.inflector" do
-      if Rails.version > "3.2"
+      if Rails.version > "3.2.x"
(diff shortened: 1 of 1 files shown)


lib/ember-cli/engine.rb
```diff
@@ -5,7 +5,7 @@ class Engine < Rails::Engine
     end
 
     initializer "ember-cli-rails.inflector" do
-      if Rails.version > "3.2"
+      if Rails.version > "3.2.x"
```
rwz: it should be `> "4"`, i think

rwz:
```ruby
"4.0.0"  > "4" # => true
"3.2.10" > "4" # => false
```

2468ben:
Thank you sir!

PR #116, opened 2015-03-18
guard-livereload issue due to symlinked dirs

jesenko:
When using guard-livereload, I noticed performance issues on OSX (100% CPU usage, livereload working unreliably). After upgrading to latest `guard` and `guard-livereload` gems, I noticed warnings related to [duplicated directory errors](https://github.com/guard/listen/wiki/Duplicate-directory-errors). This was probably the underlying cause of performance issues in the first place.

I added a note to the README.md, including the workaround by explicitly specifying directories to watch.

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,7 +143,7 @@ init` and then add the following to your `Guardfile`.
 ```ruby
 guard "livereload" do
   # ...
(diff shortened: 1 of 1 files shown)


rondale-sc:
@jesenko Thanks!  :beers:

PR #121, opened 2015-03-21
Support alternative (eg 'print') stylesheets

TeknoFiend:
In our app we need to be able to reference alternative CSS files that we will produce in Ember CLI. Specifically a print-media file. This is just one way of solving that (the simplest way that occurred to me). I just added another helper to which you can pass a list of stylesheets and options (just like the normal Rails helper). Here's how we're using my fork for testing now:

```html
    <%= include_ember_alternative_stylesheet_tags( :frontend, :print, media: 'print' ) %>
```

 I could see this being merged w/ the existing CSS helper method or maybe making that method somehow read the Ember CLI setup somehow (would still need to pass the media information somehwere). I'm happy to iterate on it or change the strategy.

### DIFF
--- a/app/helpers/ember_rails_helper.rb
+++ b/app/helpers/ember_rails_helper.rb
@@ -6,4 +6,14 @@ module EmberRailsHelper
   def include_ember_stylesheet_tags(name, options={})
     stylesheet_link_tag *EmberCLI[name].exposed_css_assets, options
   end
+
+  def include_ember_alternative_stylesheet_tags(app_name, *sources)
+    app = EmberCLI[name]
+
+    sources = sources.map do |source_or_option|
+      (source_or_option.is_a? Hash) ? source_or_option : "#{app.name}/#{source_or_option}"
+    end
+
+    stylesheet_link_tag *sources
+  end
 end


TeknoFiend:
Not compatible with `master` at the moment, I've merge the 0.2.0 branch to fix the existing/impending conflict. If necessary I can close this PR out and start over after you get 0.2.0 released.

TeknoFiend:
Updated to reflect the new location of the helpers file.

seanpdoyle:

Thanks for your patience.

seanpdoyle:
@TeknoFiend could you add some example code of how you'd picture this being used?

How are you declaring a separate `print` stylesheet with EmberCLI?

seanpdoyle:
@TeknoFiend 

We've decided that the project's future will rely heavily on `render_ember_app`:

https://github.com/thoughtbot/ember-cli-rails#configuring-the-ember-controller

That helper will renders the EmberCLI generated `index.html`. As such, the asset helpers will likely be deprecated in the future. If they're not deprecated, it's likely that they won't be extended any further.

As such, we'll have to reject this PR.

Are you relying on EmberCLI to generate your styles? 

Is it possible for your project to migrate to EmberCLI-generated styles?

If so, have you tried adding a separate stylesheet with the proper `media` attribute in your `index.html`?:

https://github.com/kellyselden/ember-cli-output/blob/89a6ef72aaa622d5e324b2b9606199a2f3b2ea12/app/index.html#L12

PR #122, opened 2015-03-21
Don't Compile Assets if `SKIP_EMBER` set

seanpdoyle:
Allows for co-located Rails / Ember apps to still utilize
`ember-cli-deploy`


rondale-sc:
LGTM.  

Do you have an example of an ember-cli-rails app using ember-cli-deploy?

PR #123, opened 2015-03-21
add note on minification

egh:
This was slowing down our build on heroku considerably. Even on my fast laptop disabling double-minification cut my asset precompilation time from 3 minutes to 1.

### DIFF
--- a/README.md
+++ b/README.md
@@ -247,6 +247,22 @@ found in both `node_modules` and `bower_components`.
 
 Now you should be ready to deploy.
 
+See also the note on [Javascript minification](#javascript-minification)
+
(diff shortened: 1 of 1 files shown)


rwz:
Much better strategy would be to disable minification on EmberCLI part. The only assets to double-minify are EmberCLI assets, so disabling Ember minification will solve this. Disabling Rails minification on the other hand will prevent ALL assets from being minified, which is undesirable.

egh:
That might be a better strategy, but it failed in my app, so it's probably good to mention both. It's also probably worth looking into the impact on source maps of double minifying.

PR #124, opened 2015-03-21
add note on minification

egh:
See #123

### DIFF
--- a/README.md
+++ b/README.md
@@ -247,6 +247,35 @@ found in both `node_modules` and `bower_components`.
 
 Now you should be ready to deploy.
 
+See also the note on [Javascript minification](#javascript-minification)
+
+## Javascript minification
+
+When precompiling assets in production, you will probably want to
+ensure that you are not minifying your javascript twice, in EmberCLI
+and in Rails. This can slow down assets precompilation considerably.
+
+There are two ways to do this. You can disable minification in Ember
+by modifying your `ember-cli-build.js` as follows:
+
+```javascript
+  var app = new EmberApp({
+    ...
+    minifyJS: false,
+    ...
+  }
+```
+
+or you can disable minification in Rails by setting:
+
+```ruby
+  config.assets.js_compressor = nil
+```
+
+in your `config/environments/production.rb` file. Either method has
+trade-offs, but you do not want to do both.
+
+
 ## Additional Information
 
 When running in the development environment, EmberCLI Rails runs `ember build`


README.md
```diff
@@ -247,6 +247,36 @@ found in both `node_modules` and `bower_components`.
...
+
+or you can disable minification in Rails by setting:
+
+```ruby
+  config.assets.js_compressor = nil
+```
+
+in your `config/environments/production.rb` file. Disabling in Ember
+is the better route if it works for you, but it may be more fragile in
+the case of javascript errors.
```
rwz: I don't get this part. How can it be more fragile? Care to elaborate?

rwz:
Other than my comment above LGTM.

README.md
egh: Your guess is as good as mine. :) I get this error in an app:

```
rake aborted!
ExecJS::ProgramError: Unexpected token: punc (:) (line: 1, col: 10, pos: 10)

Error
    at new JS_Parse_Error (/tmp/execjs20150321-24731-1hyu7atjs:2359:10623)
    at js_error (/tmp/execjs20150321-24731-1hyu7atjs:2359:10842)
    at croak (/tmp/execjs20150321-24731-1hyu7atjs:2359:19086)
    at token_error (/tmp/execjs20150321-24731-1hyu7atjs:2359:19223)
    at unexpected (/tmp/execjs20150321-24731-1hyu7atjs:2359:19311)
    at semicolon (/tmp/execjs20150321-24731-1hyu7atjs:2359:19784)
    at simple_statement (/tmp/execjs20150321-24731-1hyu7atjs:2359:22580)
    at /tmp/execjs20150321-24731-1hyu7atjs:2359:20274
    at /tmp/execjs20150321-24731-1hyu7atjs:2359:19957
    at block_ (/tmp/execjs20150321-24731-1hyu7atjs:2359:24599)
  (in /home/egh/c/tahi/tmp/ember-cli-cd98e71c-ad5a-4b81-85de-a24baaf26e75/assets/tahi/ember-data.js.map)new JS_Parse_Error ((execjs):2359:10623)
```

when I disable minifyJS in ember and use uglify in rails. I'll try to track it down, but it's pretty unclear to me.

egh:
Thanks!

README.md
rwz: Interesting. Will investigate this before merging the PR.

rondale-sc:
Disabling in ember-cli can be done automatically by the addon.

egh:
@rondale-sc That might be useful. I'm not sure if there is a one size fits all solution, though. Whether minification is better done in ember or rails might depend on the nature of your app.

Another issues is source maps. I'm still trying to get source maps to work in production with an ember-cli-rails app.

rzane:
@egh I was attempting to use the `minifyJS` option, but I was getting the same issue when running `assets:precompile`. I definitely didn't want to disable Rails asset compression. The compile was choking on ember-data.js.map.

I found this issue: https://github.com/emberjs/ember-rails/issues/428
It links to this comment: https://github.com/jamesotron/faye-rails/issues/15#issuecomment-6692493

It seems that tossing this in an initializer fixes the issue:

```ruby
Rack::Mime::MIME_TYPES.merge!(".map" => "text/plain")
```

It should probably be application/json, though. See http://stackoverflow.com/questions/18809567/what-is-the-correct-mime-type-for-min-map-javascript-source-files#18809594.

However, there's also this issue: https://github.com/sstephenson/sprockets/issues/400. It seems to suggest that map files shouldn't be asset path at all.

stevehanson:
@rzane adding the `MIME_TYPES` line in to my initializers fixed the issue for me. Thanks for sharing your solution. I'm now able to have:

```
minifyJS: {
  enabled: false
}
```

and keep the Rails JS compressor turned on. Builds are taking me about 20-30 seconds.

seanpdoyle:

We're closing this issue. If what you've described is still a problem, please comment on this issue and we'll reopen.

@rondale-sc were we able to modify https://github.com/rondale-sc/ember-cli-rails-addon to handle this automatically?

Thanks for your patience.

egh:
It is still an issue: javascript assets are being minified by ember *and* by rails asset pipeline in the default setup. This is a problem because a) it breaks sourcemaps and b) it is slow.

How to replicate: do the following (as recommended by the README):

```bash
rails new foo
cd foo
echo "gem 'ember-cli-rails', github: 'thoughtbot/ember-cli-rails'" >> Gemfile
bundle install
rails generate ember-cli:init
mkdir frontend
cd frontend
ember init
npm install --save-dev ember-cli-rails-addon@0.0.12
cd ..
```

Now, running: `RAILS_ENV=production rake assets:precompile` will minify the javascript twice: once when building with ember, and then again in the rails asset pipeline.

This behavior can be tested by suspending (`^z)` the `assets:precompile` after ember prints the temp directory where it has generated assets: (e.g., `home/egh/foo/tmp/ember-cli-e9c79eb4-c0e9-44f2-8219-66516b413c13/apps/frontend`) and copying those temp files somewhere and then comparing against the final output of the rails asset pipeline.

seanpdoyle:
@egh thanks for following up. 

Hopefully we'd be able to set this automatically through https://github.com/rondale-sc/ember-cli-rails-addon. @rondale-sc is that something that's in the works?

If not, @egh would you mind opening a PR against that project?

egh:
@seanpdoyle A decision would need to made about whether it is better for ember or the rails asset pipeline to handle this. I'm not really sure if there is a universally better choice, or if there is, which it is. But I'm happy to do what I can.

seanpdoyle:
My opinion is that it would be better for the ember addon to handle turning uglification / minification off.

Leaving it enabled Rails-side lets developers continue to minify/uglify their server-side assets.

Would you agree?

egh:
That sounds reasonable. I'll file a pull request.

egh:
It seems you were way ahead of me. :) I guess we can close this pull request, then?

Thank you for your work on this!

seanpdoyle:
@egh I've gone back on forth on where (and whether or not) minification should happen.

In the meantime, I think your PR make a lot of sense.

Would you mind updating the reference to `Brocfile.js` to `ember-cli-build.js`?

Also, I think it makes sense to omit 

> Disabling in Ember is the better route if it works for you, but it may be more fragile in the case of javascript errors.

since we really don't know which side of the pipeline is better suited for minifcation.

What do you think?

egh:
Oh, sorry, I thought https://github.com/rondale-sc/ember-cli-rails-addon/pull/16 was merged. I will update this PR.

egh:
Updated.

seanpdoyle:
Merged in ef0fc67958877c5ae37e4f1b78576c7a37e7cdac

seanpdoyle:
Thanks!

Issue #66, opened 2015-01-24
Ember app in default path degrades perf of Rails


jesenko:
When Ember app is placed in Rails `/app` folder (i.e. default ember-cli-rails path), perf of Rails (rendering pages, serving assets, etc) degrades significantly due to large number of additional files (bower_components, node_modules) in Rails `autoload` path. 

Quick profiling led to the following method call:
`Rack/ActionDispatch::Reloader/call`
* < 1ms without EmberCLI app
* ~ 150ms with EmberCLI app

As a workaround, EmberCLI app can be just placed to any non-autoload path.

To avoid perf. problems I suggest the following:
* default app EmberCLI app path should be changed to something outside Rails `app` folder, e.g. `Rails.root.join('jsapp', <appname>)`
* Possible perf. problems when placing EmberCLI app in `app` folder should be mentioned in the readme.

I have only verified this issue on my local dev machine:
Rails 4.1.7
Ruby 2.2
OS X 10.9.5

rwz:
Yep, we've noticed that as well. I wonder if there's a way to tell reloader to skip certain directories entirely. I've googled around and it seems like there're none.

I guess we'll have to change the default directory for ember apps in the next release.

rondale-sc:
@rwz I think that is probably a good idea.  Kinda sucks for peeps with existent apps.  Do you think we should display a depcrecation warning for people with cli-apps in that dir?

rwz:
Yeah, I'm thinking of the best way to handle this.

AlexCppns:
Hello, I am running into the same problem, what would be a temporary fix? puts the emberjs app in a appjs folder? I also tested ember-rails but I feel I am gonna get stuck if I dont use ember-cli from the start.

rwz:
@AlexCppns we've moved EmberCLI app into project root.

PR #126, opened 2015-03-22
Version 0.2.0

rwz:
I've rewrote some stuff and changed the way configuration works. Now everything depends on `EmberCLI.env` method that works kind of like `Rails.env` does. It controls which mode kicks in when `EmberCLI.enable!` is called.

- test: compile on first request, do nothing afterwards
- development: run build server on first request
- production: do nothing

In each mode EmberCLI assets are added to the pipeline, so `rake assets:precompile` includes it.

User can manually override which mode should be used by specifying `config.ember_cli_rails_mode = <something>` in the environment file.

We're not at 1.0 yet, so it's OK to break backward compatibility between minor versions.

@rondale-sc @seanpdoyle could you guys review changes and maybe run your projects test against this branch?

### DIFF
--- a/app/controllers/ember_tests_controller.rb
+++ b/app/controllers/ember_tests_controller.rb
@@ -18,7 +18,7 @@ class EmberTestsController < ActionController::Base
   end
 
   def app
-    EmberCLI.get_app(app_name)
+    EmberCLI[app_name]
   end
 
   def app_name
--- a/lib/ember-cli/app.rb
+++ b/lib/ember-cli/app.rb
@@ -100,14 +100,13 @@ module EmberCLI
 
     private
 
-    delegate :match_version?, :non_production?, to: Helpers
     def supported_path_method(original)
       path_method = original.to_s[/\A(.+)_path\z/, 1]
       path_method if path_method && paths.respond_to?(path_method)
     end
 
     def silence_build(&block)
-      if ENV.fetch("EMBER_CLI_RAILS_VERBOSE"){ !non_production? }
+      if ENV.fetch("EMBER_CLI_RAILS_VERBOSE"){ EmberCLI.env.production? }
         yield
       else
         silence_stream STDOUT, &block
@@ -151,7 +150,7 @@ module EmberCLI
     def check_ember_cli_version!
       version = dev_dependencies.fetch("ember-cli").split(?-).first
 
-      unless match_version?(version, EMBER_CLI_VERSION)
+      unless Helpers.match_version?(version, EMBER_CLI_VERSION)
         fail <<-MSG.strip_heredoc
           EmberCLI Rails require ember-cli NPM package version to be
           #{EMBER_CLI_VERSION} to work properly. From within your EmberCLI directory
@@ -202,7 +201,7 @@ module EmberCLI
     end
 
     def environment
-      non_production?? "development" : "production"
+      EmberCLI.env.production?? "production" : "development"
     end
 
     def package_json
--- a/lib/ember-cli/app.rb
+++ b/lib/ember-cli/app.rb
@@ -7,14 +7,13 @@ module EmberCLI
 
     class BuildError < StandardError; end
 
-    attr_reader :name, :options, :pid
+    attr_reader :name, :options, :paths, :pid
+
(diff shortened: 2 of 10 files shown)


thewatts:
@rwz -- thanks for your hard work on this! ( you as well @rondale-sc && @seanpdoyle ) !!

app/controllers/ember_tests_controller.rb
```diff
@@ -18,7 +18,7 @@ def tests_index_path
   end
 
   def app
-    EmberCLI.get_app(app_name)
+    EmberCLI[app_name]
```
seanpdoyle: :clap:

lib/ember-cli/app.rb
```diff
@@ -84,45 +82,34 @@ def wait
...
-      @ember_path ||= app_path.join("node_modules", ".bin", "ember").tap do |path|
-        fail <<-MSG.strip_heredoc unless path.executable?
-          No local ember executable found. You should run `npm install`
-          inside the #{name} app located at #{app_path}
-        MSG
+    def method_missing(method_name, *)
+      if path_method = supported_path_method(method_name)
+        paths.public_send(path_method)
+      else
+        super
```
seanpdoyle: which methods is this meant to cover?

seanpdoyle: How do you feel about something like:

```ruby
module EmberCLI
  class App < SimpleDelegator
     def initialize(name, options={})
       @name, @options = name.to_s, options
       super(PathSet.new(self))
     end
  end
end
```

instead of dynamically delegating to the paths?

lib/ember-cli/path_set.rb
```diff
@@ -0,0 +1,96 @@
...
+    end
+
+    def initialize(app)
+      @app = app
+    end
+
+    define_path :root do
+      path = app_options.fetch(:path){ default_root }
+      pathname = Pathname.new(path)
+      pathname.absolute?? pathname : Rails.root.join(path)
```
seanpdoyle: I think `Rails.root.join(pathname)` would work the same as `Rails.join(path)`?

seanpdoyle:
Could we change the rake task namespace from `ember-cli` to `ember`? 

Also, how about `ember-cli:install_dependencies` => `ember:install`?

rondale-sc:
@seanpdoyle I think that'd be a good change.  Inflecting on CLI has been a real pain.

rondale-sc:
LGTM. 

A lot of great changes and the API seems to be unbroken, unless I am missing something.  Is the minor bump because #66 or for people using internals?

lib/ember-cli/app.rb
rwz: @seanpdoyle well, it'll nehave differently. Now methods like `dist_path` delegate to `dist` method of the `paths` object. I like this `_path` suffix because it reveals the intention. I agree that all these method missing stuff could be a bit of an over engineering, but I love the result.

Far less verbose way would be something like this:

```ruby
delegate :dist, :assets, :tmp, lockfile, to: :paths, suffix: :path
```

but then you'll hav to manually mention every delegated method, and update it every time you need to introduce another one. Feels messy.

rondale-sc: FWIW I kinda like the explicitness of using delegate.  But it is a bit messy.

lib/ember-cli/path_set.rb
rwz: @seanpdoyle in this case `Rails.root.join(path)` and `Rails.root.join(pathname)` will do exactly the same. `path` is shorter :)

rwz:
> Could we change the rake task namespace from ember-cli to ember?
> Also, how about ember-cli:install_dependencies => ember:install?

Sounds good. Will add this change.

lib/ember-cli/app.rb
rwz: @rondale-sc ok, delegate is not gonna work because ActiveSupport doesn't support suffixes :(

I've send a PR adding suffix support (https://github.com/rails/rails/pull/19608), but for now we'll have to deal with messy method_missing stuff.

Issue #30, opened 2014-12-20
images and other assets


samselikoff:
where should we put assets? I have an image in my ember cli app's `public` folder, but it's not showing up in the rails app.

chrisdevor:
I created a PR that may be worth looking at regarding this issue.  It basically copies the contents of `dist/assets` into `public/assets` and then updates `manifest.json` to reference the correct fingerprinted files for `app_name.js`, `app_name.css`, `vendor.js` and `vendor.css`. Thoughts?
https://github.com/rwz/ember-cli-rails/pull/56

rondale-sc:
@chrisdevor I'm going to try to take some time tomorrow to look at #56 more closely.  :)

chrisdevor:
Sounds good.  It should work with a stand alone `rails s` environment or `rails s` in combination with `ember s --proxy:http://0.0.0.0:3000` (@ localhost:4200).  Let me know if you have any questions or issues.

chrisdevor:
Looks like [broccoli-asset-rev] (https://github.com/rickharrison/broccoli-asset-rev) has a built in [`generateRailsManifest`] (https://github.com/rickharrison/broccoli-asset-rev#options) option that I overlooked.  I'm going to investigate using that and probably push a few more fixes/updates on that branch I made over the weekend.

wsewell:
Seems like setting baseURL would not be the right thing to do since that has other effects, no?

ljuti:
Any advice on how to use `generateRailsManifest` or other means to get assets into production. I have assets, such as images, in my Rails asset pipeline that I'd like to use in Ember templates. Assets are compiled properly, but templates do not have fingerprinted URLs to them.

oliver----:
Just want to contribute a problem scenario which we are facing currently.
Unfortunately I am not experienced enough in ember-cli(-rails) to contribute to a solution.

We are using the font-awesome addon: https://github.com/lfridael/ember-cli-font-awesome
It adds its assets to "/fonts" https://github.com/lfridael/ember-cli-font-awesome/blob/master/index.js#L21-L27
But eventually these are not available in the rails app.
Any advice welcome, thanks!

rainkinz:
I had the same issue and ended up copying the fonts dir into app/assets/stylesheets. It works, but I'm sure it's not the recommended solution.

oliver----:
@rainkinz ah, interesting fix for this particular problem. Thx.

gniquil:
[...]

PR #132, opened 2015-03-27
Added the option to enable Ember CLI fingerprinting w/ bypass of asset pipeline

TeknoFiend:
Suggestion for how we can support Ember CLI fingerprints which seems necessary for referencing assets (eg images) from CSS within an Ember app. The idea here is that the optional parameter on the app definition in `config/initializers/ember.rb` causes Ember to generate fingerprinted assets. We have to copy the assets directly into `/public/assets/[APP_NAME]` so rather than linking the asset_path so that Rails does not re-fingerprint them.

In addition we fix up the Ember App's generated `manifest.json` file to prepend the app's name (e.g. `frontend/`) to the asset & file paths and merge the manifest with the manifest file in the root of `/public/assets` if it exists (in case of multiple apps). This allows the JS & CSS include helpers to work properly. This does require that `generateRailsManifest` be set to `true` in the Ember App's `Brocfile`. Perhaps we can do that automatically through the addon.

This works for multiple apps including having some apps using Ember CLI fingerprints and some using Rails fingerprinting. It does _not_ allow you to reference images within the Ember app from Rails CSS files (Rails doesn't look at the manifest file when doing asset precompilation so things work in dev but not in production). You might be able to reference images from Rails helpers in views, etc since those are dynamic and rely on the manifest file I believe.  I'm not sure that it makes sense to reach into the Ember app from Rails in any regard though I'm sure someone might find a reason to want to do so.

### DIFF
--- a/ember-cli-rails.gemspec
+++ b/ember-cli-rails.gemspec
@@ -14,4 +14,5 @@ Gem::Specification.new do |spec|
 
   spec.add_dependency "railties", ">= 3.1", "< 5"
   spec.add_dependency "sprockets", ">= 2.0"
+  spec.add_dependency "json", "~> 1.7"
 end
--- a/lib/ember-cli/app.rb
+++ b/lib/ember-cli/app.rb
@@ -1,4 +1,5 @@
 require "timeout"
+require "json"
 
 module EmberCLI
   class App
@@ -18,8 +19,12 @@ module EmberCLI
     end
 
     def compile
-      prepare
+      prepare( env: :compile )
       silence_build { exec command }
+      if use_ember_fingerprints
+        copy_fingerprinted_assets
+        fixup_asset_manifest
+      end
       check_for_build_error!
     end
 
@@ -29,14 +34,14 @@ module EmberCLI
     end
 
     def run
-      prepare
+      prepare( env: :run )
       cmd = command(watch: true)
       @pid = exec(cmd, method: :spawn)
       at_exit{ stop }
     end
 
     def run_tests
-      prepare
+      prepare( env: :test )
       tests_pass = exec("#{ember_path} test")
       exit 1 unless tests_pass
     end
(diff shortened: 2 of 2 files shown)


TeknoFiend:
Forgot to mention that this has only been tested on Rails 4. Also, in dev mode the sum link to the asset pipeline is still used since it's simpler and works.

TeknoFiend:
Looks like this does work for `image_path` and related helpers from Rails views.  But, again, they won't work from CSS/SCSS files in Rails.

sevos:
The substitution of assets paths in CSS files is quite expensive and hinders development quite a lot. In our the difference is big:

ember build  16.26s user 3.75s system 88% cpu 22.622 total
ember build --environment=production  32.54s user 2.66s system 97% cpu 36.055 total

where the difference in development (where ember serve is used) is bigger - something like 3s vs 20s, but I can not recall exactly. I would not use ember fingerprinting in development at all if not necessary.

TeknoFiend:
Yeah, you definitely shouldn't be using fingerprinting in development.

The question here is not whether or not to do substitution of asset paths in CSS. The question is do we let Ember or Rails handle the substitution. And doing it in Rails takes extra minutes not extra seconds. On a very simple test app when I enabled this fork the `assets:precompile` step in Rails took several seconds instead of 2 minutes or so.

sevos:
Thanks for the clarification.

skylar:
@TeknoFiend This looks like it might be exactly what I was looking for. Let me restate what I think this does in terms of my setup and see if we're on the same page.

I've got two Ember apps in a rails app – let's call them `frontend` and `admin` since that's a common setup. In my case, all images, css, and fonts are only in the Ember CLI app - the ember apps do not attempt to reference any rails assets. (Any assets also needed by rails are merely duplicated or sym-linked).

The primary issue then becomes, how can I specify the path to an asset (usually an image or font) in my templates and CSS so that it is compiled correctly after:

* EmberCLI fingerprints the asset
* ember-cli-rails adds a app-specific asset path to the asset (e.g. `/assets/frontend` and `/assets/admin`)
* Rails asset pipeline assembles and fingerprints its assets

I'm understand that this PR attempts to make it such that, as per the Ember CLI docs, I simply write a path to an asset as such:

   `background-image: url(/assets/images/great-logo.png);`

and after all steps above are done, the CSS or template will be modified to:

   `background-image: url(/assets/frontend/images/great-logo--777c94e23d28a0cd3827e4ac933b064f.png);`

Is that understanding correct or is my understanding clouded by my wishful thinking? :)

skylar:
Thought I don't know much about the Rails asset pipeline internals, it seems perhaps this PR only updates the manifest file for Rails, but does not rewrite CSS or template files?  In the ideal solution, the asset management ties into `broccoli-asset-rev` which ember CLI uses to update asset names.

For me, a solution has to support:

* asset namespacing in multiple ember apps (an ember-cli-rails quirk/feature)
* asset fingerprinting (ideally via Ember CLI)
* asset CDN prefixes

seanpdoyle:

Thanks for your patience.

seanpdoyle:
@TeknoFiend (and whoever else is following along in this thread) thanks for your proposed solution (and most especially your patience).

Please read (and try if possible) https://github.com/thoughtbot/ember-cli-rails/issues/30#issuecomment-154768936 and leave comments on that issue.

Thanks!

PR #140, opened 2015-04-03
Skip Middleware on `/api/` requests

seanpdoyle:
Closes https://github.com/rwz/ember-cli-rails/issues/139

Disabling middleware for API only tests keeps the feedback loop tight.
When test driving an API, this is ideal.

Ideally, this would be configurable. For now, this is a fairly sane
default.

### DIFF
--- a/lib/ember-cli/middleware.rb
+++ b/lib/ember-cli/middleware.rb
@@ -5,8 +5,10 @@ module EmberCLI
     end
 
     def call(env)
-      enable_ember_cli
-      EmberCLI.wait!
+      unless skip_middleware?
+        enable_ember_cli
+        EmberCLI.wait!
+      end
 
       if env["PATH_INFO"] == "/testem.js"
         [ 200, { "Content-Type" => "text/javascript" }, [""] ]
@@ -16,6 +18,10 @@ module EmberCLI
     end
 
     private
+
+    def skip_middleware?
+      %r{/api/}.match(env["REQUEST_URI"])
+    end
 
     def enable_ember_cli
       @enabled ||= begin


lib/ember-cli/middleware.rb
```diff
@@ -17,6 +19,10 @@ def call(env)
 
     private
 
+    def skip_middleware?
+      %r{/api/}.match(env["REQUEST_URI"])
```
seanpdoyle: @rwz @rondale-sc ideally, this would be configurable.

I envision passing a block to the middleware:

```ruby
config.use EmberCLI::Middleware do |env|
  # still not sure if this should return `true` or `false` (i.e. enable or disable)
  %r{/api/}.match(env["REQUEST_URI"]).nil?
end
```

This type of change would require being explicit, though, so we couldn't insert the middleware by default.

Does Rails gracefully handle the same middleware being declared multiple times?

seanpdoyle:
The results:

before:

```
› rspec spec/requests/api/user_spec.rb
.......

Finished in 10.22 seconds (files took 0.38471 seconds to load)
7 examples, 0 failures

Randomized with seed 61513
```

after:

```
› rspec spec/requests/api/user_spec.rb
.......

Finished in 0.45277 seconds (files took 0.37106 seconds to load)
7 examples, 0 failures

Randomized with seed 65492
```

rwz:
Well, this obviously is not generic or configurable enough to go be merged.

How about this api:

```ruby
Ember.configure do |c|
  c.app :frontend, urls: ->(url){ !url.starts_with?("/api/") }
end
```

By default each app will fire on each url (equivalent of `url: ->(*){ true }`), but user can provide any arbitrary logic to override that.

rwz:
I'm still not sure about option name, any suggestions?

seanpdoyle:
@rwz I think it depends on what the value we want to return is.

Does it make more sense to look for the match, or assert there isn't a match?

`skip` if the block is looking for a match, `enable` otherwise?

I'm in favor of `skip`

seanpdoyle:
@rwz although I do like the lambda as part of the configuration file

rwz:
I'm inclined to make it a positive match since I suspect most people has Ember app in a namespace. For instance in one of our projects Ember is served to urls that start with "/app/", so the configuration will look like like this:

```ruby
EmberCLI.configure do |c|
  c.app :frontend, enable: ->(url){ url.start_with?("/app/") }
end
```

Also, with positive match it's pretty straightforward to reverse workflow and run ember for anything that isn't "/api/", like your case.

tonywok:
@seanpdoyle I upgraded to `0.3.1` and was hoping I'd see some test speed improvements. Was enabling/disabling on api requests the reason for your drastic speed increase?

seanpdoyle:
@tonywok yes.

When TDDing API requests, precompiling the ember app doesn't make much sense. 

My ~10 seconds is thanks to skipping precompilation

Issue #139, opened 2015-04-03
Only run middleware on "ember" requests


seanpdoyle:
Could we provide a hook (either via the routes or a block in the middleware declaration) to determine if a request will be handled by the middleware?

I think the middleware is running on *all* requests. 

In testing, precompiling ember for `/api/` requests dramatically slows down an otherwise rapid feedback cycle.

What do you think? (<small>warning: pseudocode approaching</small>)

```ruby
config.use EmberCLI::ReloadMiddleware do |request|
  request.url =~ %r{/frontend/foo}
end
```

rwz:
We've tried that in the past and failed. The problem was the assets pipeline (as always).

When you're trying to request a page that has `javascript_include_tag` or `stylesheet_link_tag` in it and the files aren't actually available by the time page is rendered, rails falls back to using old-style assets paths, such as `/javascripts/frontend/frontend.js` and `/stylesheets/frontend/frontend.css`.

Because of this behavior currently we need precompile ember assets before first request hits the app.

rwz:
if you have any ideas about how to make it work, feel free to reopen this issue :)

seanpdoyle:
@rwz I'm thinking about `/api` routes that don't render pages at all.

Wouldn't a controller than calls `render json: ` work without precompiling assets?

rwz:
It would. But there's no way to tell for sure which urls require assets and which don't

seanpdoyle:
Would it make sense to allow for whitelisting URLs to skip?

seanpdoyle:
i.e. I might not know all the routes that my ember app could be served from, but it's likely that i know that `/api` or subdomain `api` won't serve asset tags

rwz:
@seanpdoyle Ok, I've created a new branch called `runner`. It supports `enable` proc option to customize which paths of your app needs certain ember app to be compiled. It also should address the issue of dying background process. Give it a spin and report if it works for you.

seanpdoyle:
![screen shot 2015-04-06 at 10 12 23 am](https://cloud.githubusercontent.com/assets/2575027/7005382/8563b270-dc45-11e4-9c85-05724bfb3f12.png)

:disappointed: could you drop the `.freeze` on the `VERSION`?

slashmili:
How does this enable options work?

I have this config for EmberCLI

    EmberCLI.configure do |c|
      c.app :myapp,
        path: "ui/myapp",
[...]

Issue #107, opened 2015-03-09
When ember command fails in development restart it.


rondale-sc:
I believe that if add a check to interrogate whether the process is still alive [here](https://github.com/rwz/ember-cli-rails/blob/master/lib/ember-cli/middleware.rb#L21) (by checking to see if the `@pid`  is still alive) we can retrigger the command or at least report it to the user.

Thoughts?

rwz:
Yeah, this is super frustrating. Will probably look into that this Friday. Totally :+1: this feature.

rwz:
@rondale-sc Added a branch called `runner` that should take care of this issue. Give it a try and report issues :)

rondale-sc:
Will do

seanpdoyle:
@rwz @rondale-sc I think this is closed by #145. Please let me know if you disagree and I'll reopen.

PR #145, opened 2015-04-06
Add Runner class to monitor and activate ember apps

seanpdoyle:
Should address #107 and #139

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,10 +51,15 @@ end
 - path - the path, where your EmberCLI applications is located. The default
   value is the name of your app in the Rails root.
 
+- enable - a lambda that accepts each requests' path. The default value is a
+  lambda that returns `true`.
+
 ```ruby
 EmberCLI.configure do |c|
   c.app :adminpanel # path is "<your-rails-root>/adminpanel"
-  c.app :frontend, path: "/path/to/your/ember-cli-app/on/disk"
+  c.app :frontend,
+    path: "/path/to/your/ember-cli-app/on/disk",
+    enable: -> path { path.starts_with?("/app/") }
 end
 ```
 
@@ -240,7 +245,7 @@ found in both `node_modules` and `bower_components`.
 
 ember-cli-rails adds your ember apps' build process to the rails asset compilation process.
 
-Now you should be ready to deploy. 
+Now you should be ready to deploy.
 
 ## Additional Information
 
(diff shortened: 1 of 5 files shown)


rwz:
Does it work for you?

Any feedback from your projects @rondale-sc?

seanpdoyle:
@rwz this worked great for me!

README.md
```diff
@@ -51,10 +51,15 @@ end
...
+- enable - a lambda that accepts each requests' path. The default value is a
+  lambda that returns `true`.
+
 ```ruby
 EmberCLI.configure do |c|
   c.app :adminpanel # path is "<your-rails-root>/adminpanel"
-  c.app :frontend, path: "/path/to/your/ember-cli-app/on/disk"
+  c.app :frontend,
+    path: "/path/to/your/ember-cli-app/on/disk",
+    enable: -> path { path.starts_with?("/app/") }
```
seanpdoyle: @rwz I've covered the positive use case, would it make sense also covering a negative case (`!path.starts_with?("/api/")`)?

rwz: I guess it's automatically implied. Anyone who sees a lambda passed in as an option should figure out it could be used for negative case by just inverting the logic.

rondale-sc:
@rwz @seanpdoyle Works on our production app without any changes!

LGTM

:beers:

(kinda obvi, but it doesn't restart ember until the next request)

rwz:
Released 0.3.0

slashmili:
How does this enable option work?

I have this config for EmberCLI

    EmberCLI.configure do |c|
      c.app :myapp,
        path: "ui/myapp",
        enable: -> path { !path.starts_with?("/api/") }
    end

And I have these routes:

    Rails.application.routes.draw do
      root 'application#index'
      scope "/api/v1" do
        resources :users
      end
    end

When I open http://localhost:3000/api/v1/users it goes to Ember and it complains that 

    Uncaught UnrecognizedURLError: /api/v1/users

seanpdoyle:
@slashmili I believe the `root` declaration must come last in your routes file.

slashmili:
Correct! Thanks

BTW it was not obvious though.

seanpdoyle:
@slashmili I agree. However, `root` isn't an ember-cli-rails construct, but something that comes from Rails itself: http://guides.rubyonrails.org/routing.html#using-root

tonywok:
I'm not super familiar with asset pipeline internals. We have some assets that are still managed by rails - (fonts, images, etc). Would it step on the ember-cli-rails middleware to also flip enabled to false on cases of `/assets/`. Looking through the source and I'm not quite sure how a "path" maps to a "request uri".

PR #148, opened 2015-04-09
explicitly register helper, fixes #147

### LINKED ISSUE #147: undefined method `include_ember_script_tags'
After upgrading to latest version of ember-cli-rails (0.3.1), ember-cli helpers are not available in views. I had to manually add `helper EmberCLI::EmberRailsHelper` to controller to include them. Probably related to https://github.com/rwz/ember-cli-rails/commit/96f4d91f27bcf2e8b2a0f78ccc6e4229abfa2b7d.

Either documentation should be updated to manual inclusion of helpers or helpers should be automatically included.

Rails version: 4.1

jesenko:
I haven't found official documentation on how to include helper to all controllers from Rails engine - this is analogous to what I found in some other github projects...

### DIFF
--- a/lib/ember-cli/engine.rb
+++ b/lib/ember-cli/engine.rb
@@ -9,5 +9,11 @@ module EmberCLI
     initializer "ember-cli-rails.enable" do
       EmberCLI.enable! unless EmberCLI.skip?
     end
+
+    initializer "ember-cli-rails.helpers" do
+      config.to_prepare do
+        ActionController::Base.send :helper, EmberRailsHelper
+      end
+    end
   end
 end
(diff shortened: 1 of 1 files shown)


lib/ember-cli/engine.rb
```diff
@@ -9,5 +9,11 @@ class Engine < Rails::Engine
     initializer "ember-cli-rails.enable" do
       EmberCLI.enable! unless EmberCLI.skip?
     end
+
+    initializer "ember-cli-rails.helpers" do
+      config.to_prepare do
+        ActionController::Base.send :helper, EmberRailsHelper
```
rwz: Seems like `helper` is not a private method, so you don't need `send` here.

jesenko: You are right, I have just committed a fix...

PR #153, opened 2015-04-14
Add disable_fingerprints option

mtmckenna:
Hello! This PR adds an option to EmberCLI.configure to disable the disabling of Ember asset fingerprinting. This is helpful for me when solving the problem of image fingerprinting as discussed in issue https://github.com/rwz/ember-cli-rails/issues/30.

In case it's useful, here's how I go about making sure the image asset fingerprints line up on disk and in source:
1. Add `extensions: ['png']` to the fingerprint hash in my Brocfile.js
2. Add `c.disable_fingerprints = Rails.env.development? || Rails.env.test?` to config/initializers/ember.rb
3. `RAILS_ENV=production rake assets:precompile` => rsync the assets directory to the server.
4. `cd frontend && ember build --environment=production` => rsync the assets/images directory to the server

It looks like there's another PR about having an asset fingerprinting option (https://github.com/rwz/ember-cli-rails/pull/132), but I'm not sure if it's trying to solve the same problem.

Let me know if you have any thoughts. Thank you!

### DIFF
--- a/lib/ember-cli/app.rb
+++ b/lib/ember-cli/app.rb
@@ -229,9 +229,13 @@ module EmberCLI
       Array.wrap(options[:exclude_ember_deps]).join(?,)
     end
 
+    def disable_fingerprints?
+      EmberCLI.configuration.disable_fingerprints
+    end
+
     def env_hash
       ENV.clone.tap do |vars|
-        vars.store "DISABLE_FINGERPRINTING", "true"
+        vars.store "DISABLE_FINGERPRINTING", disable_fingerprints?.to_s
         vars.store "EXCLUDE_EMBER_ASSETS", excluded_ember_deps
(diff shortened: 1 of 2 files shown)


TeknoFiend:
Yeah my PR (#132) was meant to address the problem of referencing images from CSS files that both live in Ember CLI. Is that what you are trying to address as well? If so maybe we should compare notes and come up with a solution that works for us both?

mtmckenna:
Yeah, let's do that! Your PR does more than mine--you're updating the manifest file. I'll give your PR a shot when I get home today and see if that works for me.

My strategy was to solve the image problem by having a custom Capistrano compile_assets task that copies over the fingerprinted images into the proper directory before uploading them to the server.

Happy to chat more about this stuff!

seanpdoyle:

We're closing this issue. If what you've described is still a problem, and #132 doesn't address it, please comment on this issue and we'll reopen.

Thanks for your patience.

PR #157, opened 2015-04-22
Fixes #156 - Issue with spaces in paths

### LINKED ISSUE #156: EmberCLI::App#command fails if there is a space in the path
I saw this while spinning up a testing suite on my company's CI server. We have long ugly path names on our Jenkins box like `/var/lib/jenkins/jobs/UXF-Services CI Pull Requests/workspace` which were causing issues like this error:
```
sh: /var/lib/jenkins/jobs/UXF-Services: Permission denied
/usr/bin/tee: Requests/workspace/log/ember-frontend.development.log: No such file or directory
```

Digging through the code a tiny bit it would make sense if the paths that come out of PathSet are escaped for spaces or any other characters that would mess with a shell command.

If I have time I'll try to throw something together for this.

Gowiem:
I've changed up the way that EmberCLI::App#exec works so that it takes it's 'command' param as an array and then uses the splat operator to deconstruct it.
This offloads the work of escaping the various different paths for spaces (which is surprisingly a pain in the ass) onto the system/spawn calls.

**Warning**: This breaks logging out to the log path alongside stdout using the 'tee' command. I don't think this should be merged until that is fixed. This issue is introduced because the pipe operator doesn't seem to be accepted by the system call when using the deconstructed arguments. I don't know why that is, and if anybody has any idea please let me know.

I'd obviously like to fix this, but I don't have the time right now and my project doesn't need the extra log file. My thoughts on fixing this is the pid handling code would need to be reworked a bit to handle a separate way of spawning off the build process such as using `IO.popen`. I'll hopefully get around to this as I don't want to get stuck on a particular version of this gem. :sweat_smile:

### DIFF
--- a/lib/ember-cli/app.rb
+++ b/lib/ember-cli/app.rb
@@ -26,8 +26,8 @@ module EmberCLI
     end
 
     def install_dependencies
-      exec "#{bundler_path} install" if gemfile_path.exist?
-      exec "#{npm_path} install"
+      exec ["#{bundler_path}", "install"] if gemfile_path.exist?
+      exec ["#{npm_path}", "install"]
     end
 
     def run
@@ -41,7 +41,7 @@ module EmberCLI
 
     def run_tests
       prepare
-      exit 1 unless exec("#{ember_path} test")
+      exit 1 unless exec(["#{ember_path}", "test"])
     end
 
     def stop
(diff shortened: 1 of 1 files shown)


Gowiem:
Doesn't seem like this is a priority for anybody so closing this out.

PR #159, opened 2015-04-23
Allow customizing precompile paths

rzane:
This allow overriding the default precompile paths for an app.

For example, if I only wanted to precompile JS files (and I do), I could do this:

```ruby
c.app :frontend, precompile: /\Afrontend\/(.*?).js$/
```

Because it's passed to Rails, you can also use a proc:

```ruby
c.app :frontend, precompile: -> { |path| path =~ /\Afrontend\// && File.extname(path).in?(['.css', '.js']) }
```

### DIFF
--- a/lib/ember-cli/app.rb
+++ b/lib/ember-cli/app.rb
@@ -193,7 +193,7 @@ module EmberCLI
     end
 
     def add_assets_to_precompile_list
-      Rails.configuration.assets.precompile << /\A#{name}\//
+      Rails.configuration.assets.precompile << options.fetch(:precompile, /\A#{name}\//)
     end
 
     def command(options={})


seanpdoyle:
@rzane thanks for this PR.

Could you provide more details about your use case? 

If you don't want to precompile stylesheets (for instance), would keeping a blank `app/styles/app.css` be sufficient?

rzane:
The reason I wanted this functionality was because including styles (even a blank file) was adding a considerable amount of time to my deploys (1 - 2 minutes). I don't know why it was taking so long, but since I didn't need the styles anyway, this seemed like an obvious solution.

seanpdoyle:
@rzane would you mind trying out:

* https://github.com/thoughtbot/ember-cli-rails/tree/remove-sprockets for `ember-cli-rails` gem
* https://github.com/rondale-sc/ember-cli-rails-addon/tree/remove-sprockets for `ember-cli-rails-addon`

Each of those branches has up-to-date README's.

rzane:
I'll try to take a look today!

seanpdoyle:
@rzane [`0.6.0` has landed](https://github.com/thoughtbot/ember-cli-rails/releases/tag/v0.6.0) and removes all sprockets integration.

I'm closing this issue. 

Please try out the new version.

PR #163, opened 2015-05-07
Fixes asset:precompile for Rails 3.1+ apps

Gowiem:
Running `rake assets:precompile` in Rails versions prior to Rails 4 [kicks off a separate rake task](https://github.com/rails/rails/blob/3-2-stable/actionpack/lib/sprockets/assets.rake#L29) (`rake assets:precompile:all`) which spins up a new environment causing the original `ember:compile` task to be wasted. Rails 3.1+ needs to hook into `rake assets:precompile:all` so it fires the `ember build` command in the **real** precompile step.

### DIFF
--- a/lib/tasks/ember-cli.rake
+++ b/lib/tasks/ember-cli.rake
@@ -15,4 +15,11 @@ namespace :ember do
   end
 end
 
-task "assets:precompile" => "ember:compile" unless EmberCLI.skip?
+unless EmberCLI.skip?
+  # Hook into assets:precompile:all for Rails 3.1+
+  if Gem::Version.new(Rails::VERSION::STRING) < Gem::Version.new("4.0.0")
+    task "assets:precompile:all" => "ember:compile"
+  else
+    task "assets:precompile" => "ember:compile"
+  end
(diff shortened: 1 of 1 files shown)


breckenedge:
:+1: This helped me a ton. Thank you.

lib/tasks/ember-cli.rake
```diff
@@ -15,4 +15,11 @@ namespace :ember do
   end
 end
 
-task "assets:precompile" => "ember:compile" unless EmberCLI.skip?
+unless EmberCLI.skip?
+  # Hook into assets:precompile:all for Rails 3.1+
+  if Gem::Version.new(Rails::VERSION::STRING) < Gem::Version.new("4.0.0")
```
rwz: You don't really need `Gem::Version` here. You can just say `Rails::VERSION::MAJOR < 4`.

seanpdoyle:
Merged in https://github.com/thoughtbot/ember-cli-rails/commit/b0b84e6d65a8009626b2ce8e437ed46706e6d304

Thanks!

PR #164, opened 2015-05-12
BUILDPACK_URL deprecated, use buildpacks:set

eliotsykes:
> Previously you could set a config var for BUILDPACK_URL, this value will still be used if set, though a buildpack value set through the CLI will take precedence. BUILDPACK_URL as a config var is now deprecated in favor of the buildpack value on the API and in the future will be migrated.

https://devcenter.heroku.com/articles/buildpacks#using-a-custom-buildpack


rwz:
:+1:

PR #172, opened 2015-06-12
Introduce `ember-cli-deploy` integration

seanpdoyle:
`EmberCLI::Deploy` takes a `namespace` (the name of your app declared in your
initializer) and handles all interaction with the Redis instance.

This is great for `staging` and `production` deploys, but introduces an extra
step in the feedback loop during development.

Luckily, `EmberCLI::Deploy` also accepts an `index_html` override, which will
replace the call to the Redis instance. This allows integration with the normal
`ember-cli-rails` workflow:

### DIFF
--- a/README.md
+++ b/README.md
@@ -200,6 +200,93 @@ watch them for changes as well.
 directories %w[app config lib spec your-appname/app]
 ```
 
+## Integrating with `ember-cli-deploy`
+
+The EmberCLI community recently unified the various deployment techniques into a
+single, core-team supported project: [ember-cli-deploy][ember-cli-deploy].
+
+This project attempts to streamline the process of pushing and serving
+EmberCLI-built static assets.
+
+To integrate with `ember-cli-deploy`'s ["Lightning Fast Deploys"][lightning]
+(using the Redis adapter), instantiate an `EmberCLI::Deploy` in your controller:
+
+```ruby
+require "ember-cli/deploy"
+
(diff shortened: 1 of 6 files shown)


seanpdoyle:
We've had some success with this approach on client and internal projects.

This approach allows us to not care about Heroku / platform-du-jour installing node / npm / ember and compiling assets, while still enabling us to keep the tight feedback loop provided by `ember-cli-rails` in development.

I'm not sure about the API, but this is how it has shaken out thusfar.

What do you think?

cc: @rwz @rondale-sc

seanpdoyle:
What do you think about this approach @rwz @rondale-sc?

Would it be better to start it as an additional gem?

aaronrenner:
@seanpdoyle This is really cool. Are you thinking about starting an additional gem? I'd be interested in helping.

seanpdoyle:
@aaronrenner that sounds great, I'd love to have some help.

Would you like to review my port of this implementation?

https://github.com/seanpdoyle/ember-cli-rails-deploy-redis/pull/1

The same invite goes out to @rwz and @rondale-sc 

I'll close this PR once I get integration working

seanpdoyle:
Closing in favor of https://github.com/seanpdoyle/ember-cli-rails-deploy-redis/releases/tag/v0.0.1

@rwz @rondale-sc should we mention that Gem in the Readme?

PR #174, opened 2015-06-19
Add a watcher option

drewchandler:
We run our app inside of Vagrant which syncs files via NFS. Unfortunately the default watcher does not detect file system changes in this setup, so we need to fallback on polling. This PR is to add support for a watcher option to the build command.

### DIFF
--- a/lib/ember-cli/app.rb
+++ b/lib/ember-cli/app.rb
@@ -126,6 +126,10 @@ module EmberCLI
       options.fetch(:build_timeout){ EmberCLI.configuration.build_timeout }
     end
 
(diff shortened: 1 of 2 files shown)


drewchandler:
@rwz any chance this could get merged or are there issues with it?

rwz:
:+1:

rwz:
Sorry, I've just shifted my attention to different stuff and ember-cli-rails is not getting enough love. :(

rwz:
Released 0.3.4

drewchandler:
Thanks!

PR #181, opened 2015-07-08
allow ember apps to bypass rails pipeline, using EmberCLI asset-rev instead

skylar:
For production deploys, there is a lot of strength in leveraging EmberCLI's use of broccoli-asset-rev. This approach also embraces the direction where Ember is the primary is front-end technology and, as such, in the primary owner of assets.

The basic approach is that the ember assets (`my-app/dist/public/assets`) are copied to the ember dist (`RAILS_ROOT/public/assets/my-app`) such that they go untouched by the Rails/Sprckets asset pipeline. The EmberCli `manifest.json` is also preserved and is tapped at runtime (after sprockets environment initialization) to update the Rails in-memory asset manifest so it can resolve URLs to traditional `asset_path` helpers to resolve paths to images or files like `my-app/my-app.js`.

Seeking to resolve issues like #30, and adapting an approach started by @TeknoFiend in #132, this PR:

* adds a new per-app flag (configured in `config/ember.rb`) `bypass_rails_asset_digests` which, when set to `true` enables this behavior
* introduces a new class `EmberCLI::AssetManifest` to load broccoli-asset-rev generated manifests, and inject this data into Sprockets manifests for both Rails 3 and Rails 4
* adds `rspec` for tests
* adds a `logger` to EmberCLI::App so we can get helpful progress messages during the asset compile, similar to what Sprockets does

I haven't tested this fully in a Rails 4 setup yet, but I wanted to get this out there for feedback. If there's interest, I'm happy to pull any fixes or adapt the PR.  Cheers.  :cocktail:


Samsinite:
@skylar, did you use
```js
  var app = new EmberApp(defaults, {
    fingerprint: {
      generateRailsManifest: true
    },
    ...
  });
```
in your `ember-cli-build.js`/`Brocfile.js` file for generating the assets or something else?

lib/ember-cli/asset_manifest.rb
```diff
@@ -0,0 +1,89 @@
...
+        assets
+      end
+    end
+
+    def exported_files
+      @data[FILES_KEY].inject({}) do |files, pair|
+        file_path,info = pair
+        info = info.merge( "logical_path" => "#{name}/#{info['logical_path']}" )
+        files["#{name}/#{file_path}"] = info
+        new_files
```
Samsinite: There is a syntax error here, should be `files` as `new_files` is undefined.

Samsinite:
Not yet working in rails 4.2, tried to use this without success even after fixing the syntax error.

skylar:
@Samsinite  yes, I used `generateRailsManifest` in the ember-cli-build.  it would be great if the add-on could set this automatically.

skylar:
@Samsinite thanks for testing in Rails 4. I fixed the tests. The Rails 4 work was more of a staring point though. The general approach seems right (injecting assets/files into the manifest) but there may be something I've missed. Since there's some interest now I'll create a sample Rails 4 project and take a look.

If there's a fix that stands out to you, lmk.

Samsinite:
I agree that the ember addon/rails gem should help setting `generateRailsManifest`. If there is interest from the folks here, it could be added as a configuration option to both the gem and addon, and be sent across to the addon at asset complication time. Another option for updating the manifest would be modifying the contents of the generated rails manifest file to mirror the ember one with the app path prefixed. Then no in-memory modifications are needed.

skylar:
@Samsinite My thought was that the in-memory merge of manifests makes more sense because other tools may need to look at the raw rails asset manifest that is generated.   Also, if rails-cli-ember is disable or not loaded in a certain environment, then you would not want the manifest changed.  For that reason, this PR "imports" the ember manifest into rails during app load, modifying the rails runtime to be ember-cli savvy (which seems to be the major approach of ember-cli-rails).

If you can post the broccoli-generated manifest and the merged manifest from your Gist, I bet we can fix up the rails 4 on the merge easily.  (i.e., it gives us a test case currently missing in the PR right now)

skylar:
@Samsinite looking at:

http://edgeguides.rubyonrails.org/asset_pipeline.html

it seems that my files index for Rails 4  was using the logical file path, not the fingerprinted path as in your gist at https://gist.github.com/Samsinite/632e99134b43e96d7162.

Do you have a sample rails4 / ember-cli app in which I can test this?

Samsinite:
I don't, I would have to put together a sample app.

Samsinite:
Also, I'm not super familiar with rails gem configuration defaults, but looks like it might need a `attr_accessor : bypass_rails_asset_digests ` in the `ember-cli-rails/lib/ember-cli/configuration.rb` file?

Samsinite:
Nevermind, I see you exposed it as an option passed to the app:
```ruby
EmberCLI.configure do |c|
  c.app :admin_panel, bypass_rails_asset_digests: true
end
```

seanpdoyle:

Thanks for your patience.

Samsinite:
Sounds awesome, thanks Sean.

seanpdoyle:
@skylar and co.

Thanks for your help. Please read https://github.com/thoughtbot/ember-cli-rails/issues/30#issuecomment-154768936 and follow up with how it works out.

seanpdoyle:
I'll be closing this PR until we sort out https://github.com/thoughtbot/ember-cli-rails/issues/30#issuecomment-154768936.

Please leave comments in that issue thread.

PR #187, opened 2015-07-28
add deployment guide for capistrano users

harisadam:
I think this would be really helpful for every Capistrano user. I spent some valuable time to debug what's wrong with my app. 

Bower is necessary for both Heroku and Capistrano.


seanpdoyle:
@harisadam Thanks for this PR, it looks like a great start!

Not being a Capistrano user, I'm not sure what most of this means. 

Would you mind expanding on this documentation with some example code snippets?

harisadam:
@seanpdoyle Well, basically it works like the config that was described for Heroku users - except the Heroku specific buildpacks -, you just need to add Bower to your ember.js app's package.json file, and Capistrano will handle everything. I think there is no need to copy an example package.json file here, ember.js users know how to install npm dependencies :)

seanpdoyle:
> ember.js users know how to install npm dependencies :)

While I would love for that to be true, everyone starts as a beginner, and I believe that code snippets are a great way to explain details to developers that are new to the Ember / JS / Node / Rails community.

What do you think about:

```
Make sure that your root level `package.json` contains `"bower"` as a development dependency:

```json
{
  "devDependencies": {
    "bower": "*"
  }
}
```
```

Or am I incorrect in suggesting the root-level `package.json`. Should it be the Ember-CLI app's `package.json` (in `frontend/package.json`, for example)?

harisadam:
ok, true, example has been added and conflicts has been fixed @seanpdoyle

seanpdoyle:
Merged in https://github.com/thoughtbot/ember-cli-rails/commit/4f3b7c00dd936b65e4824cc910b01c592a7ebf00.

Thanks!

PR #188, opened 2015-07-28
Better error messaging when Ember app is not installed.

stevehanson:
Currently, if the user tries to go to a URL in the Rails app and the Ember app has not been installed, they see an error in their browser: `EmberCLI Rails requires your Ember app to have an addon.`, even when the problem is actually that they haven't installed the app yet.

This PR adds a check before `check_addons` called `check_dependencies` that checks that the `node_modules` directory exists. If it doesn't, then the user sees a more-relevant error message:

```
EmberCLI app dependencies are not installed. From your Rails application root please run:

     $ bundle exec rake ember:install

 If you do not require Ember at this URL, you can restrict this check using the `enable`
 option in the EmberCLI initializer.
```

### DIFF
--- a/lib/ember-cli/app.rb
+++ b/lib/ember-cli/app.rb
@@ -146,6 +146,7 @@ module EmberCLI
 
     def prepare
       @prepared ||= begin
+        check_dependencies!
         check_addon!
         check_ember_cli_version!
         reset_build_error!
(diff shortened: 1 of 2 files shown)


seanpdoyle:
Merged in https://github.com/thoughtbot/ember-cli-rails/commit/cc53ec5cf71a55f4a1ec17fb21ece1a48261602d.

Thanks!

PR #196, opened 2015-08-10
Convert current environment variables into a hash before adding new variables

fusion2004:
We had the `blank_slate` gem installed which threw odd errors on attempting to clone the ENV object, due to its monkey-patching methods into Object. If we instead just create a hash with a copy of the values from ENV with `to_h`, everything seems to run perfectly fine.


rwz:
Yeah, because having an environment constant that behaves like Hash at all times, but actually is an Object is perfectly reasonable, right MRI?

PR #197, opened 2015-08-14
Allow `enable` proc to accept Rack `env`

seanpdoyle:
Some requests can be routed differently, depending on the HTTP verb.

For instance, in FormKeep, we display the Ember app for `GET /f/:token`,
but handle a Form submission via a `POST /f/:token`.


seanpdoyle:
Moved to https://github.com/rwz/ember-cli-rails/pull/217

Issue #184, opened 2015-07-16
Dont Compile Ember CLI Twice


nahtnam:
Hey!

I was chatting on the community slack and someone mentioned that ember-cli is being compiled and compressed twice. Once by npm or whatever (IDK), and then again by rails. This may be a reason why compiling/deploying the rails site take over 10 minutes for me (and other on the community slack).

Would it be possible to somehow skip the rails compilation of **only** ember-cli? I still want other js files compiled.

Thanks.

mxaly:
:+1: with two small ember apps it is painful

nahtnam:
How so?

fullofcaffeine:
Are you guys still experiencing it? It looks like it's building twice for me, too:

<img width="336" alt="screenshot_2015-08-05_19_00_05" src="https://cloud.githubusercontent.com/assets/81248/9100783/53cde276-3ba4-11e5-9b68-118176096482.png">

Is that what you're going through, too? Did you manage to solve it?

nahtnam:
Yep, and I believe that's what's taking 15 minute deploy times. I haven't
been able to find a solution yet.

On Wed, Aug 5, 2015, 5:01 PM Marcelo de Moraes Serpa <
notifications@github.com> wrote:

> Are you guys still experiencing it? It looks like it's building twice for
> me, too:
>
> [image: screenshot_2015-08-05_19_00_05]
> <https://cloud.githubusercontent.com/assets/81248/9100783/53cde276-3ba4-11e5-9b68-118176096482.png>
>
> Is that what you're going through, too? Did you manage to solve it?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/rwz/ember-cli-rails/issues/184#issuecomment-128188324>
> .
>
-- 
Thanks.

fullofcaffeine:
Yeah. I guess I will just put `ember-cli-rails` aside and use ember-cli directly. The reason I wanted to use `ember-cli-rails` was to keep things simpler at this point, but since it's a greenfield app, I don't think it's worth the effort. 

The ember app can then be directly served by `nginx` and Rails would just provide the API endpoint. I'm not really seeing any real advantages of using `ember-cli-rails` as opposed to `ember-rails`, but since `ember-cli` is *the* way to go right now when it comes to EmberJS, then `ember-rails` isn't relevant anymore, I guess.

nahtnam:
[...]

PR #199, opened 2015-08-20
Update Readme

nahtnam:
Added a way to improve build times by about 10 minutes as mentioned in #184 by @jfcouture


rwz:
:+1: thanks!

PR #217, opened 2015-10-14
Allow `enable` proc to accept Rack::Request

seanpdoyle:
Some requests can be routed differently, depending on the HTTP verb.

For instance, in FormKeep, we display the Ember app for `GET /f/:token`,
but handle a Form submission via a `POST /f/:token`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 master
 ------
 
+* Modify the `enable` configuration lambda to optionally accepts each requests'
+  `Rack::Request`. [#217]
 * Disable JS minification when generating Heroku setup [#238]
 * `BuildError#message` includes first line of backtrace. [#256]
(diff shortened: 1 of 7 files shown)


seanpdoyle:
@rondale-sc what do you think about deprecating the current `enabled: -> {}` configuration value.

We'd leave it unchanged, fire off a warning, then introduce a new config `lambda` that only accepts the request.

I'm blanking on a good name for the key... `:is_ember`, `:is_ember_request`, `:enable_build`, `:build`?

I'm not particularly fond of any of those but I hope I'm communicating my intent.

geoffharcourt:
I think `:is_ember_request` would help someone new looking at the code to understand the intent clearly.

seanpdoyle:
Closing in favor of https://github.com/thoughtbot/ember-cli-rails/pull/263

Issue #220, opened 2015-10-17
ember-cli-rails does not interact well with ember-cli addons that load external javascript


jeffreybiles:
I've been trying to use [ember-cli-stripe](https://github.com/sweettooth/ember-cli-stripe) and/or [ember-stripe-service](https://github.com/ride/ember-stripe-service), and they both are failing when using ember-cli-rails.  They both rely on loading an external javascript library from Stripe, and both are failing for the reason that the global loaded by the library (Stripe or StripeCheckout) is not defined.  When I run the ember-cli app on its own the problem disappears, leading me to believe that the problem is with ember-cli-rails.

Thanks for making this gem!  Wish I could offer a PR to go with the issue, but I'm pretty stumped.

seanpdoyle:
@jeffreybiles thanks for bringing this to our attention, it's on our radar and we're looking into it.

seanpdoyle:
@jeffreybiles initial implementation in https://github.com/thoughtbot/ember-cli-rails/pull/224

seanpdoyle:
@jeffreybiles could you give `master` a shot?

https://github.com/thoughtbot/ember-cli-rails/pull/224 should address your issues.

Also, it might be worth checking out https://github.com/thoughtbot/ember-cli-rails/pull/228

seanpdoyle:
Give https://github.com/thoughtbot/ember-cli-rails/releases/tag/v0.4.0 a shot!

jeffreybiles:
Thanks, this will be very useful for next iteration of the payment system.  If I had planned ahead better, I would have saved the spike that was using stripe within Ember.

PR #224, opened 2015-10-24
Expose EmberCLI generated index.html

seanpdoyle:
Depends on [rondale-sc/ember-cli-rails-addon#][#15].

Closes [#220].

Exposes new public methods for `EmberCLI::App`:

* `#index_html` - return the EmberCLI generated `index.html` string
* `#application_assets` - return the EmberCLI exposes app assets
* `#vendor_assets` - return the EmberCLI exposes vendor assets

Exposes new Rails helper:

* `include_ember_index_html` - Renders the `index.html` inline.
  Replaces EmberCLI asset paths with Sprockets asset paths

[#15]: rondale-sc/ember-cli-rails-addon#15
[#220]: #220

### DIFF
--- a/lib/ember-cli/app.rb
+++ b/lib/ember-cli/app.rb
@@ -1,8 +1,10 @@
 require "timeout"
+require "ember-cli/html_page"
+require "ember-cli/asset_resolver"
 
 module EmberCLI
   class App
-    ADDON_VERSION = "0.0.12"
+    ADDON_VERSION = "0.0.13"
     EMBER_CLI_VERSIONS = [ "~> 0.1.5", "~> 0.2.0", "~> 1.13" ]
 
     class BuildError < StandardError; end
@@ -21,6 +23,7 @@ module EmberCLI
         prepare
         silence_build{ exec command }
         check_for_build_error!
+        copy_index_html_file
         true
       end
     end
@@ -36,6 +39,7 @@ module EmberCLI
       cmd = command(watch: true)
       @pid = exec(cmd, method: :spawn)
       Process.detach pid
+      copy_index_html_file
       set_on_exit_callback
     end
 
@@ -49,12 +53,30 @@ module EmberCLI
       @pid = nil
     end
 
-    def exposed_js_assets
-      %W[#{name}/vendor #{name}/#{ember_app_name}]
+    def index_html(sprockets)
+      asset_resolver = AssetResolver.new(
+        app: self,
+        sprockets: sprockets,
+      )
+      html_page = HtmlPage.new(
+        asset_resolver: asset_resolver,
+        content: index_file.read,
+      )
+
+      html_page.render
     end
 
-    def exposed_css_assets
-      %W[#{name}/vendor #{name}/#{ember_app_name}]
+    def exposed_js_assets
+      [vendor_assets, application_assets]
+    end
+    alias exposed_css_assets exposed_js_assets
+
+    def vendor_assets
+      "#{name}/vendor"
+    end
+
+    def application_assets
+      "#{name}/#{ember_app_name}"
     end
 
     def wait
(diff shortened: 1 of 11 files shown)


seanpdoyle:
@rondale-sc this implements what we discussed Friday morning.

lib/ember-cli/app.rb
```diff
@@ -237,7 +277,17 @@ def environment
...
+        JSON.parse(package_json_file_path.read).with_indifferent_access
+    end
+
+    def addon_package_json
+      @addon_package_json ||=
+        JSON.parse(addon_package_json_file_path.read).with_indifferent_access
+    end
+
+    def addon_version
+      addon_package_json.fetch("version", "0.0.0")
```
tute: Why do we need to specify the default value here? The user might have not defined it?

lib/ember-cli/asset_resolver.rb
```diff
@@ -0,0 +1,53 @@
+module EmberCLI
+  class AssetResolver
+    def initialize(options)
```
tute: Can we use keyword arguments here? `def initialize(app:, sprockrets:)`

lib/ember-cli/html_page.rb
```diff
@@ -0,0 +1,12 @@
+module EmberCLI
+  class HtmlPage
+    def initialize(options)
```
tute: Same note about keyword arguments.

spec/ember-cli/asset_resolver_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+    HTML
+  end
+
+  class SprocketsMock
+    def asset_path(routes, type: :javascript)
+      extension = if type == :javascript
+                    "js"
+                  else
+                    "css"
+                  end
```
tute: Style: for short conditionals like this I personally favor ternaries: `extension = type == :javascript ? "js" : "css"`. I realize this is not a popular opinion in my teams. :)

lib/ember-cli/asset_resolver.rb
seanpdoyle: Unfortunately, there are Rails 3.1 users we'd like to support.

Am I too pessimistic assuming that if users are tied up with Rails 3.1 their Ruby versions won't be up to date?

Our [support matrix](https://github.com/thoughtbot/ember-cli-rails/blob/master/.travis.yml#L8-L11) tests against 2.0 -> 2.2, even though that commit is only a few days old.

What are your opinions on library support for older rubies?

lib/ember-cli/app.rb
seanpdoyle: That's a good point, the addon is either there and valid, or it's not.

lib/ember-cli/asset_resolver.rb
tute: > What are your opinions on library support for older rubies?

Older rubies, your take! Depends much on our users indeed.

For Rails it is easier: Rails 3.1 and 4.0 are not supported, and have known vulnerabilities. I'd recommend dropping them, and users can keep using previous ember-cli-rails releases. Before the release of v1 I'd consider dropping 3.2 too, as it's close to real EOL as well.

seanpdoyle: I agree, support for older Rails versions should be dropped.

The same could probably be said for Ruby versions.

Since I've only recently inherited maintainership of this project, I'll try not to rock the boat.

In the future I'll make some sort of announcement specifying version boundaries.

tute: You can follow along in https://rubygems.org/gems/ember-cli-rails to check downloads per version. Most people are on latest 0.3.5. While we are on `0.` we have freedom to decide on dependencies, if we need and wnat to.

PR #225, opened 2015-10-25
Add Happy Path Smoke Test

seanpdoyle:
Embed a Rails application in `spec/dummy`.

Configure `capybara-webkit` for CI according to old [suspenders]
configuration.

Embed an EmberCLI application in `spec/dummy/my-app`. Exclude from
version control. When running `bin/setup`, a generated EmberCLI will be
cloned from [kellyselden/ember-cli-output].

[suspenders]: thoughtbot/suspenders@ac3924d
[kellyselden/ember-cli-output]: https://github.com/kellyselden/ember-cli-output


bin/setup
```diff
@@ -0,0 +1,26 @@
...
+  npm install -g bower
+fi
+
+if ! [ -d spec/dummy/my-app ]; then
+  git clone https://github.com/kellyselden/ember-cli-output.git spec/dummy/my-app
+fi
+
+root="$(pwd)"
+
+cd ${root}/spec/dummy/my-app &&
```
teoljungberg: I feel like replacing `${root}` with `$(pwd)` would make this easier to read, what do you think?

spec/dummy/app/controllers/application_controller.rb
```diff
@@ -0,0 +1,5 @@
+class ApplicationController < ActionController::Base
```
teoljungberg: I think there are templates for rails, which is one file as an entire rails application. I know @derekprior have mentioned it at some point, but I can't find it now.

derekprior: https://github.com/rails/rails/tree/master/guides/bug_report_templates

I'm not sure there's one that would be appropriate here or not.

bin/setup
seanpdoyle: I tried that:

```
bin/setup: line 24: cd: /Users/seanpdoyle/src/ember-cli-rails/spec/dummy/my-app/spec/dummy: No such file or directory
```

I think it's important to capture the current directory for the sake of the `cd` commands, right?

spec/dummy/app/controllers/application_controller.rb
seanpdoyle: I've stolen the majority of [spec/dummy/application.rb](https://github.com/thoughtbot/ember-cli-rails/blob/a6584138ebc37a5b029d5a7a723ad4362384bd0f/spec/dummy/application.rb) from [Clearance](https://github.com/thoughtbot/clearance/blob/7a63a2a9363729c743693d74f725337b860ee314/spec/dummy/application.rb)

seanpdoyle: @teoljungberg thanks for that link, though. I really like the idea of `bundler/inline`, and `gemfile { # ... }`.

I'll try those in the future when testing against multiple versions of Rails

seanpdoyle:
@derekprior @teoljungberg does this look good to merge?

derekprior:
:+1:

Issue #151, opened 2015-04-13
Avoid adding acronyms to ActiveSupport::Inflector


CloCkWeRX:
Due to https://github.com/rails/rails/issues/17879 and https://github.com/rwz/ember-cli-rails/blob/master/lib/ember-cli/engine.rb#L5 we started seeing addresses in our system render as

```
123 Example Street, North CLI Ffe WA
```
Instead of
```
123 Example Street, NORTHCLIFFE WA
```
... when we used .titleize or similar.

The fact that ember CLI rails had defined something that changes global state in this way entirely non-intuitive; at least when you expect this to be a self contained library. This should be something that users of the gem opt into; or contained to only the code that requires it - IE; introducing a wrapper function that is aware of the few places this is needed would be a better approach.

seanpdoyle:
This behavior and global state leakage is unacceptable. We'll look into this.

Thank you for bringing this to our attention.

seanpdoyle:
@CloCkWeRX would you mind verifying a potential fix for this issue?

Please update your `Gemfile`:

```rb
# Gemfile
gem "ember-cli-rails", github: "thoughtbot/ember-cli-rails", branch: "sd-remove-inflection"
```

Then try to reproduce the improperly inflected title.

seanpdoyle:
@CloCkWeRX I've merged that branch, could you retry with `master`:

```rb
# Gemfile
gem "ember-cli-rails", github: "thoughtbot/ember-cli-rails"
```

CloCkWeRX:
Seems happy.

PR #233, opened 2015-11-02
Remove `"CLI"` inflection

seanpdoyle:
Closes [#151].

> Due to [rails/rails#17879] and (the [inflection] definition) we
> started seeing addresses in our system render as
>

```
123 Example Street, North CLI Ffe WA
```

> Instead of

```
123 Example Street, NORTHCLIFFE WA
```
> when we used `.titleize` or similar.

To fix this, we'll use the `EmberCli` module in our implementation. To
ensure backward compatibility, we'll alias the `EmberCli` module to
`EmberCLI`.

[inflection]: https://github.com/rwz/ember-cli-rails/blob/master/lib/ember-cli/engine.rb#L5
[#151]: https://github.com/thoughtbot/ember-cli-rails/issues/151
[rails/rails#17879]: https://github.com/rails/rails/issues/17879

### DIFF
--- a/app/controllers/ember_tests_controller.rb
+++ b/app/controllers/ember_tests_controller.rb
@@ -18,7 +18,7 @@ class EmberTestsController < ActionController::Base
   end
 
   def app
-    EmberCLI[app_name]
+    EmberCli[app_name]
   end
 
   def app_name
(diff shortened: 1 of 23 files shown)


tute:
Tricky! LGTM.

PR #235, opened 2015-11-02
Change seconds to milliseconds in readme

ariroseonline:


seanpdoyle:
@ariroseonline the `build_timeout` value is, in fact, in seconds:

https://github.com/thoughtbot/ember-cli-rails/blob/444449ef40d43130b8262d1b4ea21929d2de9c5b/lib/ember-cli/app.rb#L100

See the [Timeout.timeout](http://ruby-doc.org/stdlib-1.9.3/libdoc/timeout/rdoc/Timeout.html#module-Timeout-label-Synopsis) documentation.

seanpdoyle:
Thanks for your contribution. 

If you need to pass in thousands of seconds, there might be other issues with how your project is configured.

PR #236, opened 2015-11-03
Fail the process if `exec` fails

seanpdoyle:
When running `install_dependencies` or `compile`, failures won't halt
the host program.

This could lead to false positives and difficult to debug scenarios.

Consumers would prefer this to fail as early on as possible.

### DIFF
--- a/lib/ember-cli/app.rb
+++ b/lib/ember-cli/app.rb
@@ -60,7 +60,8 @@ module EmberCli
 
     def run_tests
       prepare
-      exit 1 unless exec("#{ember_path} test")
+
+      exec("#{ember_path} test")
     end
 
     def stop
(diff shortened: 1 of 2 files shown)


lib/ember-cli/app.rb
```diff
@@ -339,7 +339,7 @@ def env_hash
 
     def exec(cmd, method: :system)
       Dir.chdir root do
-        Kernel.public_send(method, env_hash, cmd, err: :out)
+        Kernel.public_send(method, env_hash, cmd, err: :out) or fail(cmd)
```
geoffharcourt: I think you might also be able to use `Kernel.exit` here: https://github.com/thoughtbot/parity/blob/256f9a3f995d21fa6059efcd523855f2a1a0c381/bin/production#L8

seanpdoyle: @geoffharcourt yup, ended up using that [here](b8b8963e3e754dc52b8c2992dbbff19d8cb498b4)

geoffharcourt:
This slipped my mind before, but is there a way to test this by simulating the failure of a dependency build?

seanpdoyle:
@geoffharcourt do you mean in a test?

geoffharcourt:
Yeah. Unfortunately we didn't do it in Parity on the same type of change, so maybe that's overkill.

PR #238, opened 2015-11-03
Turn off JS Minification for Heroku

seanpdoyle:
Given the outcome of [#123] and [#124], we don't take a stand on where
minification should happen.

However, to make sure that Heroku deployments don't take too long, the
generator can disable Rails-side minification.

The benefit of doing this in the generator (as opposed to automatically
in [the addon]) is that the change will introduce diff noise, which is
explicit and easy to back out of.

[#123]: https://github.com/thoughtbot/ember-cli-rails/pull/123
[#124]: https://github.com/thoughtbot/ember-cli-rails/pull/124
[the addon]: https://github.com/rondale-sc/ember-cli-rails-addon/blob/2f2a47efe6e81a2bcd43ec9ff0f89fc18bed5a03/index.js#L25-L32

### DIFF
--- a/lib/generators/ember-cli/heroku/heroku_generator.rb
+++ b/lib/generators/ember-cli/heroku/heroku_generator.rb
@@ -13,6 +13,16 @@ module EmberCli
       run "chmod a+x bin/heroku_install"
     end
 
+    def config_js_compressor
+      production_config = "config/environments/production.rb"
+
+      inject_into_file production_config, before: "end\n" do
+        <<-RUBY
+  config.assets.js_compressor = nil
+        RUBY
+      end
+    end
+
     def inject_12factor_gem
       gem "rails_12factor", group: [:staging, :production]
     end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 master
 ------
 
+* Disable JS minification when generating Heroku setup [#238]
 * `BuildError#message` includes first line of backtrace. [#256]
 * Exposes `build_timeout` configuration as `ENV["EMBER_BUILD_TIMEOUT"]`.
 * Change default `build_timeout` to `15` seconds.
@@ -9,6 +10,7 @@ master
 * `manifest.json`. Since we now defer to EmberCLI, we no longer need to
   manually resolve asset URLs. [#250]
 
+[#238]: https://github.com/thoughtbot/ember-cli-rails/pull/238
 [#256]: https://github.com/thoughtbot/ember-cli-rails/pull/256
 [#250]: https://github.com/thoughtbot/ember-cli-rails/pull/250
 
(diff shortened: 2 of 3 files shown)


lib/generators/ember-cli/heroku/heroku_generator.rb
```diff
@@ -13,6 +13,15 @@ def copy_setup_heroku_file
       run "chmod a+x bin/heroku_install"
     end
 
+    def config_js_compressor
+      production_config = "config/environments/production.rb"
+
+      inject_into_file production_config, before: "end\n" do <<-RUBY
```
geoffharcourt: Oh, Rubocop. So eager and so wrong here.

geoffharcourt:
@seanpdoyle to be clear, this is only disabling the Rails-side minification, correct? JS is still being minified by Ember?

After reading #123 that was my impression, but wanted to confirm my understanding.

drapergeek:
What happens if I'm using Ember for part of my app but using jquery and a few other things for another area of my app? Isn't this eliminating the minification there?

seanpdoyle:
@geoffharcourt that's correct.

There are PRs (https://github.com/rondale-sc/ember-cli-rails-addon/pull/12, https://github.com/rondale-sc/ember-cli-rails-addon/pull/16) that disable minification on the addon side of things, but they do it automatically and implicitly, without giving the consumer the chances to make their decision.

There still a big decision to be made here, but I'd like to test it out on Heroku first to see how it goes.

seanpdoyle:
@drapergeek yes, it is.

Fortunately, since this is a generator, there will be diff noise.

That diff noise could be reverted, and at that point you could manually set `minifyJS: false` in your Ember app, or do nothing and minify both assets twice.

seanpdoyle:
@drapergeek Ideally, on a greenfield app (like the one Geoff and I are working) you'd keep your assets on the frontend.

drapergeek:
>@drapergeek Ideally, on a greenfield app (like the one Geoff and I are working) you'd keep your assets on the frontend.

That's an ideal situation. There are definitely cases where you want things to be split.

>Fortunately, since this is a generator, there will be diff noise.

That's fine for our case but I worry that others won't see what is happening and they lose speed due to an oversight. Not sure how to rectify that issue while still fixing this problem though.

geoffharcourt:
@seanpdoyle I think a change to the README to clarify would be helpful, but the code side of this looks good.

seanpdoyle:
@drapergeek would you prefer disabling minification in EmberCLI -- relying on the Asset Pipeline for minification?

drapergeek:
>@drapergeek would you prefer disabling minification in EmberCLI -- relying on the Asset Pipeline for minification?

Definitely not. Of the two options, I significantly prefer this.

This really doesn't affect me since I deploy with ember-cli-deploy anyway. I'd leave on minification for our other stuff. I'm just concerned about others who don't notice what is happening.

PR #242, opened 2015-11-04
Upgrade hero background and add Github link to navigation

jasonramirez:
screen shot 2015-11-04 at 5 17 00 pm


seanpdoyle:
:heart_eyes:

_sass/_hero.scss
```diff
@@ -1,26 +1,35 @@
 .hero {
   @include size(100vw 100vh);
   align-items: center;
-  background-color: $ember-orange;
+  background-image: url("/images/hero-background-triangles.svg");
```
joshuaogle: Does `image-url()` not work here?

joshuaogle:
I may be missing something, but why would `images/hero-background.jpg` not just simply be a CSS gradient?

Overall this looks great! :+1:

jasonramirez:
@joshuaogle great catch! The `hero-background.jpg` image is not actually being used. I went the way of the gradient, and forgot to delete this.

PR #246, opened 2015-11-06
Simplify Test Pages

seanpdoyle:
Dedicate each feature test configuration its own [HighVoltage] route and
view.

Decreases clutter in test pages and allows new tests to occupy their own
view.

Additionally, we must modify the `my-app` configuration to use the
`hash` router, so that our JS won't fail depending on the URL of the
route being loaded.

[HighVoltage]: https://github.com/thoughtbot/high_voltage


Gemfile
```diff
@@ -6,6 +6,7 @@ gem "rails", "4.2.4"
 gem "pry"
 
 group :development, :test do
+  gem "high_voltage", "~> 2.4.0"
```
geoffharcourt: This is an extremely elegant approach. :+1:

bin/setup
```diff
@@ -1,7 +1,11 @@
-#!/bin/sh
+#!/usr/bin/env bash
```
geoffharcourt: Was there something not `sh`-compatible that drove this change?

geoffharcourt:
You've already merged, but looks good. One question about the hashbang change.

bin/setup
seanpdoyle: It complained about `function()`

PR #253, opened 2015-11-09
Document `rake ember:test`

seanpdoyle:
The gem exposes the `ember:test` Rake task to execute Ember's test suite.

Without proper documentation, it isn't clear that extending a host
application's test suite to execute `ember:test` is a Good Idea :tm:.

### DIFF
--- a/README.md
+++ b/README.md
@@ -263,6 +263,20 @@ example, `/ember-tests`) and the name of the Ember app.
 For example, to view tests of the `frontend` app, visit
 `http://localhost:3000/ember-tests/frontend`.
 
+### Integrating with Rake
+
+EmberCLI Rails exposes the `ember:test` Rake task to execute Ember's test suite.
+
+If you're using Rake to run your test suite, make sure to configure your test
+task to depend on `ember:test`.
+
+For example, to configure a bare `rake` command to run both RSpec and Ember test
+suites, configure the `default` task to depend on both `spec` and `ember:test`.
+
+```rb
+task default: [:spec, "ember:test"]
+```
+
 ## Serving from multi-process servers in development
 
 If you're using a multi-process server ([Puma], [Unicorn], etc.) in development,


geoffharcourt:
LGTM. Do you want to automatically append this task to the default rake task when an app is built?

seanpdoyle:
@geoffharcourt that would be great.

Is there a good, test framework agnostic way to do that?

Ideally, we'd still have `rake spec` ONLY run RSpec, and `rake test` ONLY run minitest.

Presumably we'd introspect the pre-defined Rake tasks and append `ember:test` onto `default`, right?

geoffharcourt:
This SO answer implies you can declare it when you declare your subtask, but I'm wondering if there's a better-looking strategy.

geoffharcourt:
Also relevant: http://blog.jayfields.com/2008/02/rake-task-overwriting.html

mike-burns:
I'd be surprised by @geoffharcourt 's suggested behavior, in a couple of places:

- If I jumped onto a client project that worked like this and I wanted to only run a subset of the tests, I'd have to dig rather deep to discover that this gem hijacked `:default`.
- If my tests, bundler, or Spring were acting up, I'd want the ability to debug with everything in front of me instead of scattered throughout loaded gems.
- If I added this gem to my Gemfile, I'd expect to find some documentation about how to enable it for the default rake task. I'd be surprised that this modifies my default task, which would lead to me reading through the code for this gem to see what other surprises are in store, and I'd add it to the software matrix with a note like "Note: this modifies your :default task."

Additionally, if this overrides the :default task, will you then have to provide a way to override that for people who want it done differently?

geoffharcourt:
Overriding a rake task is a pretty annoying thing to do as specified in the blog post above. @mike-burns you've sold me that documentation is enough.

seanpdoyle:
@mike-burns @geoffharcourt thanks for the helpful discussion.

Now that we've decided that documentation is the best approach, what do we think about the proposed changes to the README?

mike-burns:
LGTM.

benoittgt:
Thanks for this PR.

PR #263, opened 2015-11-13
Introduce `mount_ember_app` route helper

seanpdoyle:
Abstract away the details of routing to an ember application.

To configure Rails to route requests to the `frontend` Ember
application:

```rb

Rails.application.routes.draw do
  mount_ember_app :frontend, to: "/"
end
```

Ember requests will be set `params[:ember_app]` to the name of the application.
In the above example, `params[:ember_app] == :frontend`.

**Routing options**

* `path` - The path to handle as an Ember application. This will only apply to
  `format: :html` requests. Additionally, this will handle child routes as well.
  For instance, mounting `mount_ember_app :frontend, to: "/frontend"` will handle a
  `format: :html` request to `/frontend/posts`.
* `controller` - Defaults to `"ember_cli/ember"`
* `action` - Defaults to `"index"`

You should now be able to boot your Rails application, navigate to the `root`
route, and see your EmberCLI app!

By default, routes defined by `ember_app` will be rendered with the internal
`EmberCli::EmberController`. The `EmberCli::EmberController` renders the Ember
application's `index.html` and injects the Rails-generated CSRF tags into the
`<head>`.

To override this behavior, specify the `controller` and `action` options:

```rb

Rails.application.routes.draw do
  mount_ember_app :frontend, to: "/", controller: "application", action: "index"
end
```

To serve the EmberCLI generated `index.html`, use the `include_ember_index_html`
helper in your view:

```erb
<!-- app/views/application/index.html.erb -->
<%= include_ember_index_html :frontend %>
```

To inject markup into page, pass in a block that accepts the `head`, and
(optionally) the `body`:

```erb
<!-- app/views/application/index.html.erb -->
<%= include_ember_index_html :frontend do |head| %>
  <% head.append do %>
    <%= csrf_meta_tags %>
  <% end %>
<% end %>
```

When serving the EmberCLI generated `index.html`, don't use Rails' layout HTML:

```rb
class ApplicationController < ActionController::Base
  def index
    render layout: false
  end
end
```

In addition to rendering the EmberCLI generated `index.html`, you can inject the
`<script>` and `<link>` tags into your Rails generated views:

```erb
<!-- app/views/application/index.html.erb -->
<%= include_ember_script_tags :frontend %>
<%= include_ember_stylesheet_tags :frontend %>

### DIFF
--- a/lib/ember-cli/engine.rb
+++ b/lib/ember-cli/engine.rb
@@ -1,12 +1,23 @@
 module EmberCli
   class Engine < Rails::Engine
+    initializer "ember-cli-rails.rendering" do
+      require "ember-cli/ember_controller"
+      require "ember-cli/route_helpers"
+    end
+
     initializer "ember-cli-rails.enable" do
       EmberCli.enable! unless EmberCli.skip?
     end
 
-    initializer "ember-cli-rails.helpers" do
-      config.to_prepare do
-        ActionController::Base.helper EmberRailsHelper
+    config.to_prepare do
+      ActionController::Base.helper EmberRailsHelper
+    end
+
+    config.after_initialize do
+      if defined?(ApplicationController)
+        require "ember-cli/controller_extension"
+
+        ApplicationController.include(EmberCli::ControllerExtension)
       end
     end
   end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,9 @@
 master
 ------
 
+* Introduce `mount_ember_app` route helper [#262]
 * Remove support for viewing Ember tests through Rails. Instead, use `ember
-  test` or `ember test --serve` from within the Ember directory.
+  test` or `ember test --serve` from within the Ember directory. [#264]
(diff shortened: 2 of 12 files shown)

### CODE REVIEW

File: lib/ember-cli/engine.rb [hunk 1]
```diff
@@ -1,12 +1,27 @@
...
-      config.to_prepare do
-        ActionController::Base.helper EmberRailsHelper
+    config.to_prepare do
+      ActionController::Base.helper EmberRailsHelper
+    end
+
+    config.after_initialize do
+      if defined?(ApplicationController)
+        ApplicationController.class_eval do
+          before_action do
```
derekprior: Would it make sense for this to live in a module that you `ApplicationController.extend(module_name)` here?

seanpdoyle: I forgot how to Ruby, thank you

derekprior: Also, I don't know much about this project, but would you consider having the user include this module/concern on their own? Maybe generating the include as part of the install generator?

PR #300, opened 2015-11-23
Use `posix-spawn` instead of `Kernel`

seanpdoyle:
The [posix-spawn] gem removes process forking overhead by using special
process spawning interfaces (`posix_spawn()`, `vfork()`, etc.).

> This gem can keep your application's heap from being copied when
> forking command line processes

[posix-spawn]: https://github.com/rtomayko/posix-spawn

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 master
 ------
 
+* Use `POSIX::Spawn::{spawn,system}`.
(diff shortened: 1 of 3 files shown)


seanpdoyle:
Closing, since JRuby support prevents this from landing.

https://github.com/rtomayko/posix-spawn/pull/41#issuecomment-89401569

Issue #304, opened 2015-11-23
Assets return 404 after ~5 minutes on dev


botandrose:
Assets come through at first, but ~5 minutes later they return 404s. A rails server restart fixes this for another 5 minutes.

botandrose:
Interesting, looks like the directory gets deleted when the server is up for five minutes:

```
$ ls -la tmp/ | grep ember-cli-
drwxr-xr-x   4 micah micah 4096 Nov 23 16:14 ember-cli-c49aa0b4-61e6-4dbc-b7e4-6a38669eeafb
$ sleep 300
$ ls -la tmp/ | grep ember-cli-
$
```

Struggling to come up with a hypothesis for this behavior...

botandrose:
The plot thickens. Directory does not disappear when running `rails s`, but does with apache 2.4 + latest passenger (5.0.21)! Closing for now, as this appears to be Passenger's fault.

botandrose:
Okay, turns out this is fallout from using EmberCLI-Rails in development with Passenger's default configuration. Specifically, "smart spawning" mode, and the preloader's timeout of five minutes.

Passenger's smart spawning boots the app in a separate process (aka preloader), then subsequently forks a new worker processes off of the preloader. The benefit of this is that spinning up new worker processes is very fast, and all the worker processes get to share memory via copy-on-write. These worker processes are the ones that actually serve responses, while the preloader never does. If the preloader goes five minutes without spawning a new process, it will shut down to conserve memory.

So, when a Rails app with EmberCLI-Rails boots in this default Passenger environment, the `at_exit` hook<sup>1</sup> is registered in the _preloader_ process. Then, when that process dies after the timeout, the `at_exit` hook is run, which eventually deletes the `tmp/ember-cli-rails-[uuid]` directory<sup>2</sup>, and we get our 404s from the still-running _worker_ processes.

There are very easy workarounds one could employ in the Passenger configuration, like disabling smart spawning, or setting the preloader timeout to zero. However, this took me a fair bit of effort to even figure out what was going on, much less strategies for working around it, so I think it'd be nice to try to address this in EmberCLI-Rails itself, if possible. The first idea that comes to mind is deleting old apps on boot, rather than deleting the current app on shutdown. What do you think? I'd be happy to put together a PR for this, if this strategy sounds acceptable.
____
Footnotes:
[...]

PR #308, opened 2015-11-24
Delete old compiled apps on boot, instead of on exit.

botandrose:
This works better with forking multi-process servers. Specifically, Phusion
Passenger with its default configuration would delete the current compiled app
prematurely, yielding 404s.

Closes [#304].

[#304]:
https://github.com/thoughtbot/ember-cli-rails/issues/304

### DIFF
--- a/lib/ember-cli-rails.rb
+++ b/lib/ember-cli-rails.rb
@@ -32,7 +32,7 @@ module EmberCli
   def enable!
     @enabled ||= begin
       Rails.configuration.assets.paths << root.join("assets").to_s
-      at_exit{ cleanup }
+      cleanup
       true
     end
   end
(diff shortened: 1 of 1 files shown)


botandrose:
@seanpdoyle WDYT?

The glob for old compiled apps is a little naive, but I think its probably good enough. If you prefer, I can add a UUID regex to ensure we're only deleting the ember-cli-rails app directories.

Also, there are no explicit tests for this behavior change, but there weren't any for deleting old apps before, and the feature test smokes it, at least. Let me know if you want me to write some unit tests.

seanpdoyle:
Merged as part of https://github.com/thoughtbot/ember-cli-rails/commit/2d51daa8cbcd51d24bf94f6d509c476cbdafc5fe

seanpdoyle:
Thanks!

botandrose:
:tada:

PR #313, opened 2015-11-29
Prevent generating git repo for Ember application

alexlafroscia:
Following the original documentation, I accidentally created a Git repo for the Ember application inside the existing Git repo for the Rails app.  However, if you specify `--skip-git` when you create the Ember app, this can be avoided.


seanpdoyle:
Merged in https://github.com/thoughtbot/ember-cli-rails/commit/7badc2581e0158bb8017cf5a24a39ea8d50b77df

Thanks!

PR #317, opened 2015-12-02
Silence exec output on success; display on failure

botandrose:
Motivation: For a reason outside the scope of this PR, `rake ember:compile` was silently failing with a 1 exit status... no output at all. Dug through the source to find that `Kernel.system`'s options were set to redirect a output to a log file. Let's show that output when the command fails!

Two changes here, by commit:

1. Turns out `#silence_build` wasn't really doing anything, since we're already redirecting the output in the `Kernel.system` options. However, it will silence my attempts at printing the output failure! It -- along with the `EMBER_CLI_RAILS_VERBOSE` env flag -- was undocumented, and not used anywhere else. The delete key is my favorite key.

2. On failure, have `#exec` print the failing command and its output to STDERR for easier debugging. So yeah, the verbage I've put in there is pretty crappy; you may want to change the exact output.

Again, no tests for this :frowning:,  but `EmberCli::Shell` appears to be mostly untested (only vicariously and minimally via `EmberCli::App`), so ¯\\\_(ツ)\_/¯

WDYT?

### DIFF
--- a/lib/ember_cli/shell.rb
+++ b/lib/ember_cli/shell.rb
@@ -54,7 +54,15 @@ module EmberCli
     end
 
     def exec(command)
-      Kernel.system(env, command, process_options) || exit(1)
+      Kernel.system(env, command, process_options) ||
+        exit_with_debug_info(command)
+    end
+
+    def exit_with_debug_info(command)
+      STDERR.puts "command has failed: #{command}"
+      STDERR.puts "command output:"
+      STDERR.puts File.read(paths.log)
+      exit 1
     end
 
     def process_options
--- a/lib/ember_cli/shell.rb
+++ b/lib/ember_cli/shell.rb
@@ -13,7 +13,7 @@ module EmberCli
     end
 
     def compile
-      silence_build { exec ember.build }
+      exec ember.build
     end
 
     def build_and_watch
@@ -76,14 +76,6 @@ module EmberCli
 
     def detach
       Process.detach pid
-    end
-
-    def silence_build(&block)
-      if ENV.fetch("EMBER_CLI_RAILS_VERBOSE") { EmberCli.env.production? }
-        yield
-      else
-        silence_stream(STDOUT, &block)
-      end
     end
   end
 end
--- a/lib/ember_cli/shell.rb
+++ b/lib/ember_cli/shell.rb
@@ -1,4 +1,5 @@
 require "ember_cli/command"
+require "open3"
 
 module EmberCli
   class Shell
@@ -54,14 +55,18 @@ module EmberCli
     end
 
     def exec(command)
-      Kernel.system(env, command, process_options) ||
-        exit_with_debug_info(command)
+      options = process_options.dup
+      out = options.delete(:out)
+      combined_output, exit_status = Open3.capture2e(env, command, options)
+      File.write out, combined_output
+      exit_with_debug_info(command, combined_output) unless exit_status.success?
+      true
     end
 
-    def exit_with_debug_info(command)
+    def exit_with_debug_info(command, combined_output)
       STDERR.puts "command has failed: #{command}"
       STDERR.puts "command output:"
-      STDERR.puts File.read(paths.log)
+      STDERR.puts combined_output
       exit 1
     end
 
--- a/spec/lib/ember_cli/shell_spec.rb
+++ b/spec/lib/ember_cli/shell_spec.rb
@@ -1,0 +1,64 @@
+require "ember-cli-rails"
+require "tmpdir"
+
+describe EmberCli::Shell do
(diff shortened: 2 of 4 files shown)


seanpdoyle:
@botandrose thanks for this PR!

I understand where your coming from and agree with the spirit of this PR, however, there are some details that will have to change.

I agree, `#compile` should not silently swallow errors, since the only time it would be called is during a `rake` invocation (during deploys or local debugging). I am alright with removing `#silence_build`.

Unfortunately, I disagree with logging the error file to STDERR and exiting with a status of `1` for `#exec` and `#spawn`.

`#spawn` is only ever invoked in development to spin up the `ember build --watch` process that backs the Ember app. Logging errors to the log file is sufficient, since the log files serves as inter-process communication between the Rails and Ember CLI processes.

`ember-cli-rails` [watches the log file](https://github.com/thoughtbot/ember-cli-rails/blob/7badc2581e0158bb8017cf5a24a39ea8d50b77df/lib/ember_cli/build_monitor.rb#L38-L54) for errors [piped from the ember-cli-rails-addon](https://github.com/rondale-sc/ember-cli-rails-addon/blob/983637954d30af09f0eb38412a910159e82c21a7/index.js#L29-L31).

`#exec` is used in [a few places](https://github.com/thoughtbot/ember-cli-rails/blob/8eec101fe0b882653db89894221918390465de57/lib/ember_cli/shell.rb#L34-L45) other than `#compile`. `#exec` pipes STDOUT to the log file.

I don't think piping the contents of the log file to STDERR would be useful, since it'd probably be a bit too noisy.

What do you think about piping `#exec`'s STDERR to `paths. build_error_file`, then replaces the `exit(1)` with raising a [`BuildError`](https://github.com/thoughtbot/ember-cli-rails/blob/8eec101fe0b882653db89894221918390465de57/lib/ember_cli/errors.rb#L2) containing the contents of the build error file? (This would only make sense for `#exec`, rolling back the changes to `#spawn`)

botandrose:
Roger that. Thanks for the feedback! I figured you'd have superior insight on exactly how to implement this... I'm still getting familiar with the codebase. I'll restore `#spawn`'s original implementation, presently, but I think there's still some discussion needed on the error logging.

I considered raising an exception, but decided against it, thinking that the ruby stack trace would not be helpful... just extra noise. The way I see it, if a command fails, we're probably only interested in knowing what that command was, and seeing its failing output. Seeing that it came from `#exit_with_debug_info`, which came from `#exec`, and so on inside of EmberCLI-Rails is probably not relevant, unless you're developing EmberCLI-Rails itself.

Concerning how we handle the error output, just capturing and displaying STDERR isn't sufficient, unfortunately, because `ember build`, at least, is sending its errors over STDOUT. We could pipe them both together into a single log, and display that?

Lastly, at least for the error I was seeing (a build error), it turns out that displaying the log file isn't too noisy, because each `#exec` logging stomps on the previous `#exec`'s log, so we only see output for the failing command. Which is exactly what I want in that situation, at least.

What do you think?

botandrose:
Correction: subsequent calls `#exec` do not stomp on previous logs. It just appeared that way because the previous commands were not outputting anything to STDOUT. The log is probably being cleared on boot, too, or something. Will have to think some more about the strategy of dumping the log to screen, at least.

botandrose:
Correcting the correction. I ran some tests, and believe my original hypothesis is correct. Subsequent calls to `#exec` does indeed appear to stomp on previous logs.

botandrose:
@seanpdoyle Okay, I've pushed some more commits, this time with tests to clarify the desired behavior. I've combined STDOUT and STDERR into one stream and am pushing that to the log, then conditionally displaying it if the command fails. Let me know what you think about this approach.

botandrose:
Okay, I think this is looking pretty good now. I've got it working with both dumping directly to STDERR, and with raising an exception. I've enclosed screenshots of both below to illustrate the difference. Let me know which one you'd prefer, and we'll go with that.

screenshot from 2015-12-02 16 03 01

screenshot from 2015-12-02 16 02 54

seanpdoyle:
@botandrose thanks for putting in the work on this!

I'm leaning toward logging to STDERR.

I won't have much time this week to dedicate to this, but I'm very interested in getting a working solution in the future.

botandrose:
@seanpdoyle Great! If you're okay with logging to stderr, then I think the PR is ready to go. Let me know if there's anything else I can do to help.

seanpdoyle:
@botandrose thanks for all the hard work!

I'm closing this PR in favor of https://github.com/thoughtbot/ember-cli-rails/pull/324.

Would you mind: 

1. pulling that branch down and making sure it behaves the way you'd expect
1. weighing in on that PR

PR #324, opened 2015-12-03
Improve command failure reporting

seanpdoyle:
* Remove `#silence_build` method from `#exec.`
* Tell the user what happened when a command fails.
* Print both the command and error message for easier debugging.
* Capture both STDOUT and STDERR in #exec.

Commands may output important debugging information via either channel.
Use `Open3.capture2e` instead of `Kernel.system`. It appears to be
impossible to capture both `STDERR` and `STDOUT` simultaneously with the
latter.

### DIFF
--- a/spec/lib/ember_cli/runner_spec.rb
+++ b/spec/lib/ember_cli/runner_spec.rb
@@ -1,0 +1,34 @@
+require "ember_cli/runner"
+
+describe EmberCli::Runner do
+  describe "#run!" do
+    context "when the command fails" do
+      it "writes STDERR and STDOUT to `err`" do
+        out = StringIO.new
+        err = StringIO.new
+        runner = EmberCli::Runner.new(err: err, out: out)
+
+        expect { runner.run!("echo 'out'; echo 'err' > /dev/stderr; exit 1") }.
+          to raise_error(SystemExit)
+
+        [err, out].each(&:rewind)
+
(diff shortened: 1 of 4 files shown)


spec/lib/ember_cli/runner_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+
+describe EmberCli::Runner do
+  describe "#run!" do
+    context "when the command fails" do
+      it "writes STDERR and STDOUT to `err`" do
+        out = StringIO.new
+        err = StringIO.new
+        runner = EmberCli::Runner.new(err: err, out: out)
+
+        expect { runner.run!("echo 'hello'; exit 1") }.
```
botandrose: This test isn't currently enforcing the capture of the command's STDERR, as stated in `it`. Perhaps something like:

```ruby
runner.run!("echo 'hello'; echo 'world' > /dev/stderr; exit 1")
...
expect(err.read).to match(/hello\nworld/)
expect(out.read).to eq("hello\nworld\n")
```

botandrose:
Other than the note above, looks good to me! I'm glad you refactored away `process_options`, btw, my version was pretty unfortunate, with the duping and deleting, etc.

seanpdoyle:
@botandrose :green_apple:

botandrose:
:+1: Works great, thank you!

PR #347, opened 2015-12-10
Invoke `EmberCli[ember_app].build` from helpers

seanpdoyle:
While the initial invocation can be slow, subsequent calls end up being
no-ops.

This change address a slew of errors and user-filed issues about assets
not being compiled before requests are served.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 master
 ------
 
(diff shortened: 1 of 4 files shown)


fusion2004:
:+1: Awesome!

botandrose:
:+1: Thank you!

PR #349, opened 2015-12-10
[BUGFIX] Ensure Ember build is complete.

### LINKED ISSUE #331: No such file or directory @ rb_sysopen - /Users/collins/git/<app>/tmp/ember-cli-<uuid>/apps/frontend/index.html
I am getting this error on startup in development environment, in cases where the first request to Rails is *not* one for my mounted Ember app.

    Errno::ENOENT in Assets#index
    No such file or directory @ rb_sysopen - /Users/collins/git/<app>/tmp/ember-cli-<uuid>/apps/frontend/index.html

When this happens I must restart my local development server and make sure the first request is to `/`, where my app is mounted, so it can trigger ember-cli-rails.

I have a series of endpoints mounted under `/api` separate from any Ember app, so I've been using those endpoints to test.

Steps I take to reproduce:

- start up my local development server with latest `ember-cli-rails` version 0.5.8
- don't request anything in a web browser
- run `curl -s http://localhost:8123/api/assets`
- request http://localhost:8123/ in a web browser
- get the above error

patcoll:
Checks to see if index.html file exists for each app before declaring the build complete. Fixes #331.

### DIFF
--- a/lib/ember_cli/build_monitor.rb
+++ b/lib/ember_cli/build_monitor.rb
@@ -32,7 +32,7 @@ module EmberCli
     attr_reader :name, :paths
 
     def complete?
-      !paths.lockfile.exist?
+      !paths.lockfile.exist? && paths.dist.join("index.html").exist?
     end
 
     def build_error?


lib/ember_cli/build_monitor.rb
```diff
@@ -32,7 +32,7 @@ def wait!
     attr_reader :name, :paths
 
     def complete?
-      !paths.lockfile.exist?
+      !paths.lockfile.exist? && paths.dist.join("index.html").exist?
```
seanpdoyle: Does this fix the bug you've been seeing?

The addon creates the lockfile when the build starts, then removes it when it's finished.

If the lockfile no longer exists, I'm not sure what good waiting on the `index.html` file will do.

It's likely that it'll hang infinitely

patcoll: Yes, it does fix my issue.

I couldn't find where the lockfile gets removed... I looked but no luck. More correct behavior would be to wait until index.html exists in dist before the lockfile gets removed. 

Overall, expected behavior is that index.html should exist before declaring build done though, right? Open to suggestions on where to ensure that logic.

seanpdoyle: @patcoll the lockfile is removed by https://github.com/rondale-sc/ember-cli-rails-addon/blob/cadb98ad22462664fda501fb7213649123fd4c63/index.js#L63-L65.

I agree, the build should block until `index.html` exists.

seanpdoyle: @patcoll one thing I couldn't clarify before the end of our session:

Are you using `rack-timeout` or have any other request timeout strategies?

patcoll: Yes, I do. My rack-timeout logic looks like this (in a Rails initializer):

    if Rails.env.development? or Rails.env.test?
      Rack::Timeout.timeout = 60
    else
      Rack::Timeout.timeout = 15  # seconds
    end

With the newest non-Sprockets builds it never reaches anywhere near 15 seconds of initial load time. It's more like 5 seconds on average right now.

patcoll: Hm, interesting. Could be a fsExtra.copySync bug or similar.

seanpdoyle: The addon no longer copies the `index.html` file:

https://github.com/rondale-sc/ember-cli-rails-addon/commit/fd6d6bac28758d6700012bd12d02d1d25a4c9156

Copying the index was a means of integrating with Sprockets, and is no longer necessary.

patcoll:
I'll use these two versions and see if that removal of fs-extra fixes my issue:

- https://github.com/thoughtbot/ember-cli-rails/commit/b4a055e8
- https://github.com/rondale-sc/ember-cli-rails-addon/commit/72ac7aa

patcoll:
Nope. It took 5-6 tries to get the error, but I'm still getting it. Same one as in the gist.

patcoll:
Definitely seems like a race condition of some sort.

seanpdoyle:
@patcoll https://github.com/rondale-sc/ember-cli-rails-addon/pull/27

patcoll:
That totally works. You're a savant. I'm closing this.

Issue #319, opened 2015-12-02
error deploying via capistrano


dchersey:
Hi again -- I've just built a new Ubuntu 14 server and am getting an error from ember-cli-rails when deploying via capistrano. 

```
DEBUG [23492c3d] Command: cd /var/www/releases/20151202203519 && ( RBENV_ROOT=~/.rbenv RBENV_VERSION=2.2.3 NODE_VERSION=v4.2.1 RAILS_ENV=production RBENV_ROOT=~/.rbenv RBENV_VERSION=2.2.3 ~/.rbenv/bin/rbenv exec bundle exec rake assets:precompile )
DEBUG [23492c3d] 	rake aborted!
DEBUG [23492c3d] 	TypeError: no implicit conversion of nil into String
DEBUG [23492c3d] 	/var/www/shared/bundle/ruby/2.2.0/bundler/gems/ember-cli-rails-7badc2581e01/lib/ember_cli/path_set.rb:83:in `initialize'
DEBUG [23492c3d] 	/var/www/shared/bundle/ruby/2.2.0/bundler/gems/ember-cli-rails-7badc2581e01/lib/ember_cli/path_set.rb:83:in `Pathname'
DEBUG [23492c3d] 	/var/www/shared/bundle/ruby/2.2.0/bundler/gems/ember-cli-rails-7badc2581e01/lib/ember_cli/path_set.rb:83:in `block in bower'
DEBUG [23492c3d] 	/var/www/shared/bundle/ruby/2.2.0/bundler/gems/ember-cli-rails-7badc2581e01/lib/ember_cli/path_set.rb:82:in `tap'
DEBUG [23492c3d] 	/var/www/shared/bundle/ruby/2.2.0/bundler/gems/ember-cli-rails-7badc2581e01/lib/ember_cli/path_set.rb:82:in `bower'
DEBUG [23492c3d] 	/var/www/shared/bundle/ruby/2.2.0/bundler/gems/ember-cli-rails-7badc2581e01/lib/ember_cli/shell.rb:40:in `install'
DEBUG [23492c3d] 	/var/www/shared/bundle/ruby/2.2.0/bundler/gems/ember-cli-rails-7badc2581e01/lib/ember_cli/app.rb:52:in `install_dependencies'
[...]

PR #350, opened 2015-12-10
including instructions for Capistrano deployment in README.md

dchersey:
See issue #319

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,6 +163,19 @@ applications into the project.
 
 [buildpack]: https://devcenter.heroku.com/articles/using-multiple-buildpacks-for-an-app#adding-a-buildpack
 
+## Capistrano
+EmberCLI-Rails executes both `npm install` and `bower install` as part of the asset precompilation step of a Capistrano build. This requires that the correct version of these executables be found in the path of your deployment SSH session. Since Capistrano executes in a very basic shell, it is not sufficient to place the PATH in your `.bash_profile`. 
+
+Adding the bin directory of your node install to your capistrano deploy script will enable EmberCLI-Rails to find the correct bower and npm executables.
+
+```ruby
+#config/deploy/production.rb
+set :default_env, {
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -163,6 +163,19 @@ applications into the project.
...
 
+## Capistrano
+EmberCLI-Rails executes both `npm install` and `bower install` as part of the asset precompilation step of a Capistrano build. This requires that the correct version of these executables be found in the path of your deployment SSH session. Since Capistrano executes in a very basic shell, it is not sufficient to place the PATH in your `.bash_profile`. 
+
+Adding the bin directory of your node install to your capistrano deploy script will enable EmberCLI-Rails to find the correct bower and npm executables.
+
+```ruby
+#config/deploy/production.rb
+set :default_env, {
+  'PATH' => '/home/deploy/.nvm/versions/node/v4.2.1/bin:$PATH' #note: using NVM to set node version in this example. Replace with output of 'which npm' in your server environment.
```
seanpdoyle: > Replace with output of 'which npm' in your server environment.

will the output of `which npm` be swappable with `/home/deploy/.nvm/versions/node/v4.2.1/`? 

Is there another command that will produce the proper path?

dchersey: It depends on your node configuration. If you are using the system node, ‘which npm’ returns /usr/bin/npm.  If you are using NVM node version manager, it will return the bin directory from the in-use version (as in my example). 

You need to remove /npm from the answer to ‘which npm’ to get the correct directory for the path.  And bower needs to be installed under that version of node as well. 

I agree this may not be the clearest explanation — perhaps saying: 

1) Make sure you have selected the production version of Node that your server uses (i.e. via nvm use xxx, or just installing it as system).
2) Execute npm install -g bower 
3) Take the path from these (i.e. ‘which bower’ or ‘which npm’) and make sure they are first in the PATH used by capistrano, below.


What do you think?

seanpdoyle:
merged in https://github.com/thoughtbot/ember-cli-rails/commit/c8c1fd17a77aa2314a7d744d6252ffce7f123c27.

Thanks @dchersey!

Issue #311, opened 2015-11-26
404 for frontend.js with include_ember_script_tags :frontend


sevab:
I'm loading my ember app only inside some of my rails views via `include_ember_script_tags :frontend`. However, upon starting the server I get a 404 for `frontend.js`.

The issue is solved by a hack where I temporarily add `mount_ember_app :frontend, to: "/"` on top of `routes.rb` and reload the page, which presumably triggers ember to build the assets. Upon first reload I get a blank page because the assets are obviously not built yet, and upon the second reload I get a blank page because the route is caught by `mount_ember_app :frontend, to: "/"`, which tries and fails to load `tmp/../apps/frontend/index.html`. But now that assets have been built, removing `mount_ember_app` from routes lets me load my intended rails view as well as `frontend.js`.

Any changes to ember app's files after this successfully trigger rebuilding of the asset, but not before the hack has been made.

The issue has only appeared after updating from v0.3.5 to 0.5.6.

seanpdoyle:
Thanks for opening this issue and bringing this to our attention!

This issue sounds like a duplicate of #298, so I'm closing it, since it's already on our radar.

As a potential quick-fix (if a fix is time-sensitive), try this out:

https://github.com/thoughtbot/ember-cli-rails/issues/298#issuecomment-159352737

sevab:
As I understand `sd-manifest` has now been merged into master https://github.com/rondale-sc/ember-cli-rails-addon/pull/23, so I have updated the gem to `0.5.8` and `ember-cli-rails-addon 0.5.2`, but the issue remains.

sevab:
Worth noting that I haven't actually tested `sd-manifest` back when the issue occurred, so I'm not sure whether it fixed my issue in the first place.

seanpdoyle:
@sevab those changes have probably already been removed in the latest release.

I'm working on releasing `0.6.x` with some breaking changes that simplifies everything.

I'll also be updating an example repository that will hopefully clarify how the library is to be used.

Updates are forthcoming...

sevab:
@seanpdoyle thank you!

seanpdoyle:
@sevab would you mind trying out:

* https://github.com/thoughtbot/ember-cli-rails/tree/remove-sprockets for `ember-cli-rails` gem
* https://github.com/rondale-sc/ember-cli-rails-addon/tree/remove-sprockets for `ember-cli-rails-addon`

Each of those branches has up-to-date README's.

[...]

PR #352, opened 2015-12-11
Introduce `mount_ember_assets` routing helper

seanpdoyle:
Inspired by [#311].

The `mount_ember_app` helper mounts both a `Rack::File` application that
serves the EmberCLI-generated assets, as well as a catchall route (scoped
to the `to` argument) that redirects all requests to the Ember
application.

For users that have unconventional applications with Ember code embedded
in Rails pages, this is unwanted behavior.

The `mount_ember_assets` only mounts routes to serve the
EmberCLI-generated assets.

[#311]: https://github.com/thoughtbot/ember-cli-rails/issues/311#issuecomment-163739834

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 master
 ------
 
+* Introduce `mount_ember_assets` to serve Ember assets in asset-helper style
+  projects.
+
 0.6.0
 -----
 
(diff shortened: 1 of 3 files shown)


sevab:
Just tested `sd-mount-ember-assets` and everything works. However, mounting assets on `mount_ember_assets :frontend, to: '/'`, and including assets on, say, `my/stuff` doesn't seem to require `<base href="/my/stuff">`, and actually only works with `<base href="/">`. Not sure if this is a bug, but it is certainly more convenient.

botandrose:
:+1: Works for my use-case, too. Can finally upgrade from 0.4! Thank you.

PR #357, opened 2015-12-14
Fix JRuby failures in CI

seanpdoyle:
CI is failing for JRuby.

This commit gets the test suite passing.


seanpdoyle:
There must've been an issue with Circle / JRuby.

The branches are now passing without code changes.

PR #367, opened 2015-12-20
Replace character-literal declarations with string

seanpdoyle:
http://stackoverflow.com/a/16641934

Using `?;` instead of `";"` obscures intent and isn't doing us any
favors in terms of performance, since the literal is represented as a
string.

This commit removes all occurrences.


seanpdoyle:
merged in https://github.com/thoughtbot/ember-cli-rails/commit/f1d32e6a244f8249bd2f6f7eead0c85810a5f6f8

PR #371, opened 2015-12-23
Boot Ember processes on initialization

seanpdoyle:
Avoid delay for first page load by starting Ember builds during server
startup.

Since controllers no longer invoke `Ember.build`, we can simplify the
tests' controllers, deferring logic to the default `HighVoltage`
controllers.


seanpdoyle:
Closing this, since it affects the boot time for other `rails` commands, like `rails console`.

PR #379, opened 2016-01-11
Fixed heroku generator when deploying project with multiple apps

jeanbono:
The current heroku_install bin wasn't able to deploy multiple apps at the same time.
The cd inside the loop was working for the first app since we were inside the project root but the second one was failing because it was trying to cd from the directory of the first app.

### DIFF
--- a/lib/generators/ember/heroku/templates/bin_heroku_install.erb
+++ b/lib/generators/ember/heroku/templates/bin_heroku_install.erb
@@ -3,11 +3,10 @@
 set -e
 
 bower="$(pwd)/node_modules/.bin/bower"
-project_root="$(pwd)"
 
 for app in <%= app_paths.map { |app_path| %{"#{app_path}"} }.join(" ") -%>; do
   cd $app &&
     npm install &&
-    $bower install &&
-    cd $project_root
+    $bower install
+  cd -
 done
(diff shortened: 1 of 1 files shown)


lib/generators/ember/heroku/templates/bin_heroku_install.erb
```diff
@@ -3,9 +3,11 @@
...
 
 bower="$(pwd)/node_modules/.bin/bower"
+project_root="$(pwd)"
 
 for app in <%= app_paths.map { |app_path| %{"#{app_path}"} }.join(" ") -%>; do
   cd $app &&
     npm install &&
-    $bower install
+    $bower install &&
+    cd $project_root
```
seanpdoyle: Thanks for proposing this change.

What do you think about this instead:

```bash
   cd $app &&
      npm install &&
     $bower install

  cd -
done
```

jeanbono:
Oh sure, that was just for clarity sake.

jeanbono:
@seanpdoyle Updated my PR with your proposition.

seanpdoyle:
Merged in https://github.com/thoughtbot/ember-cli-rails/commit/0482a4650af688291b40f1debf6091fb8db3e457.

Thanks!

PR #387, opened 2016-01-15
Move dependencies from Gemfiles to gemspec

seanpdoyle:
When adding new dependencies, adding to a single gemspec as a
development dependency is DRYer than adding them to each file in
`gemfiles/*`.


seanpdoyle:
Closing in favor of https://github.com/thoughtbot/ember-cli-rails/pull/563.

Issue #373, opened 2015-12-26
Docs: inheriting the default controller and adding before-filters


nruth:
Great work on reviving this gem, in particular `render_ember_app`!

I'd like to suggest documenting how to add before-filters without duplicating the controller implementation. 

For example, the following adds Devise authentication before sending the index page (security through obscurity for the JS file URLs, which may be desirable for admin applications).

```ruby
# routes, assuming authentication already set up
mount_ember_app :admin_panel, to: "/", controller: 'authenticated'

# app/controllers/authenticated_controller.rb
class AuthenticatedController < EmberCli::EmberController
  before_action :authenticate_admin_user!
end
```

If this looks ok I'll make a PR for the readme and add a test if there isn't one.
I'm asking first because I think this is adding something to the public API?

seanpdoyle:
Thanks for opening this issue!
Could you elaborate on your use case -- I don't understand the reasoning behind extending `EmberCli::EmberController` (instead of `ApplicationController`).

[`render layout: false`](https://github.com/thoughtbot/ember-cli-rails/blob/1d06d46a822c608958027a836e5815295898b1de/app/controller/ember_cli/ember_controller.rb#L4) is the only meaningful behavior, and [that's already documented](https://github.com/thoughtbot/ember-cli-rails/blob/1d06d46a822c608958027a836e5815295898b1de/README.md#overriding-the-controller).

nruth:
Thanks for reading! I don't mind if this gets rejected, just trying to give something back.

I was originally going to suggest a readme/wiki example for authentication or adding controller before-filters, but noticed that this way of extending the controller seems more like the root cause of why I found it fiddly to set up. Improving that (if it does) seemed more valuable than making the readme longer.

It's part laziness and part future-proofing, or perhaps just personal preference. I think it simplifies set-up where mostly default behaviour is needed.

This is all a bit tenuous because there aren't that many concepts or lines of code involved.

#### Use case: authentication / adding controller callbacks / filters

My use case is adding authentication (to protect the app source), by inserting a Devise before-filter for one of the two Ember apps hosted by this gem. For instance :frontend served to everyone, but :admin_panel only served to logged-in admins.

#### Maintenance

This can be done following the readme, but it means duplicating two files from the gem: the controller and the view. That's fine until they change in the gem, and the copies in the app get left behind?

Since the controller is very simple and the view template uses `render_ember_app ` this seems like a moot point.

#### Simplicity

[...]

PR #389, opened 2016-01-19
docs: before-filters by subclassing EmberController #373

nruth:
Following on from #373 this adds a quick way to plug in before-filters to the readme.

Example use cases:
- Analytics: hit mixpanel on download, including user session data
- Authentication: only serve admin app javascript files to logged-in admins
- Rollout / Flipper: only serve prototype app code to beta-testers

This is already possible and works. The issue is whether it belongs in the readme. 

I think it's nice, because it took me a while to piece together how to add a few before-filters, and it adds only the new controller to the application. 

On the other hand, it's more text, and there might be some reason not to encourage people to extend the controller?

If it's accepted I'll add a test (feature spec) before it's merged.

### DIFF
--- a/README.md
+++ b/README.md
@@ -357,6 +357,21 @@ class ApplicationController < ActionController::Base
 end
 ```
 
+#### Adding before-filters
+
+To simply add before-filters to the default controller, you can subclass the `EmberCli::EmberController`. This alone won't secure your system, but gives a quick way to hook up things like rollout, flipper, or authentication checks if you need to.
+
+```rb
+# routes, assuming authentication already set up
+mount_ember_app :admin_panel, to: "/admin", controller: 'admin_frontend'
+
+# app/controllers/authenticated_controller.rb
+class AdminFrontendController < EmberCli::EmberController
+  before_action :authenticate_admin_user!
+end
+```
+
+
 ### Rendering the EmberCLI generated JS and CSS
 
 Rendering EmberCLI applications with `render_ember_app` is the recommended,


seanpdoyle:
@nruth thanks for this PR!

You're right, there are a slew of use cases for `before_filter` calls.

> Authentication: only serve admin app javascript files to logged-in admins

I do this on almost all of my `ember-cli-rails` applications.

I am in favor of this PR in spirit, but I'm not confident that encouraging users to extend `EmberCli::Controller` is a good thing.

I'll experiement with improving the current README on my own. We'll revisit this PR afterwards.

nruth:
> I am in favor of this PR in spirit, but I'm not confident that encouraging users to extend EmberCli::Controller is a good thing.

That's the bit I'm not sure about too. It works, but I guess makes it harder to change the internals.

Another way could be having a rails generator that spits out the necessary controller and view files quickly, but it may be overkill as they're quite small?

seanpdoyle:
@nruth that's certainly a possible solution -- we do that for some of Clearance and Administrate.

I'm not sure if that's a better approach, since it increases the potential for their code being out of sync with `ember-cli-rails`' expectations.

nruth:
Yes that's why I liked subclassing. 

For keeping in synch / expected it's nice to get deprecation warnings.
I've never worked on a library where I've set that up, but I imagine something like this could work:

```rb
class AdminFrontendController < ApplicationController
  before_action :authenticate_admin_user!
  EmberCli.acts_as_ember_controller
 end
```

Which just imports a module defining the default show action.
It could be deprecated and replaced in a new release if the behaviour was very different.

PR #391, opened 2016-01-21
Fix nil error when running rake ember:install

krishandley:
```
TypeError: no implicit conversion of nil into String
```

Ruby 2.3.0
Node v5.0.0
ember-cli-rails 0.7.0

I didn't have bower installed, so adding path.to_s allowed the code to tell me I needed to install bower.


seanpdoyle:
@krishandley merged in https://github.com/thoughtbot/ember-cli-rails/commit/ce5db79326f5aac72615dcb257d10e02bdc1aeb8.

Thanks!

Issue #392, opened 2016-01-25
No `ember-cli` executable found for `frontend`. Install it:


rhysdavies1994:
This is my first time setting up an app with EmberJS and the ember-cli-rails gem.
I've followed the steps in the readme for this git, but when i go to my index page i get the following error:

No `ember-cli` executable found for `frontend`. Install it:
it tells me to run npm install, which changes nothing.

I'm working on Windows 7.

Has anyone else had this issue?

seanpdoyle:
@rhysdavies1994 thanks for opening this issue!

Where is the `frontend` directory relative to your Rails app?

Does `frontend/package.json` declare `"ember-cli"` as a dependency?

If so, what version of `"ember-cli"` is declared as a dependency?

Does `frontend/node_modules/.bin/ember-cli` exist? If so, is it executable?

rhysdavies1994:
@seanpdoyle 
the frontend is stored in the root directory of my rails app. e.g my rails folder is called "learning-ember" generated from "rails new learning-ember". the frontend folder is "learning-ember/frontend".

From what I can see it looks to declare "ember-cli" : "1.13.15" under devDependencies

There is a bunch of ember-cli folders in "frontend/node_modules", although inside of .bin there is only "ember" and "ember.cmd"

I've attached a few files. An image of the error, full trace and my package.json file to assist you.

[ember-cli-rails error.png](https://cloud.githubusercontent.com/assets/7247102/12553073/4a626b28-c3b0-11e5-8ab8-a94811e47c84.PNG)
[ember-cli-rails-full-stack.txt](https://github.com/thoughtbot/ember-cli-rails/files/103303/ember-cli-rails-full-stack.txt)
[package.json.txt](https://github.com/thoughtbot/ember-cli-rails/files/103302/package.json.txt)

seanpdoyle:
> There is a bunch of ember-cli folders in "frontend/node_modules", although inside of .bin there is only "ember" and "ember.cmd"

My mistake -- `ember` is the name we're looking for. There is both an `ember.cmd` and `ember` version?

patcoll:
This seems to happen to me as well.

[...]

PR #395, opened 2016-01-29
Resolve `ember` executable with full path

seanpdoyle:
Closes [#392].

Resolve `ember` executable with full path within `node_modules`, instead
of depending on the presence of `node_modules/.bin`.

[#392]: #392


staylorwr:
:clap: Thanks @seanpdoyle

Issue #394, opened 2016-01-29
ember:install should blow away deps and reinstall from scratch if ember version says so?


botandrose:
Ember CLI ships with [ember-cli-dependency-checker](https://github.com/quaertym/ember-cli-dependency-checker), which provides the `ember version` command, which is akin to `bundle check`. I have enhanced the `ember:install` rake task to `rm -rf node_modules` before installing if `ember version` returns non-zero.

This seems like something that might belong in Ember CLI Rails itself. WDYT? Should I prepare a PR?

seanpdoyle:
@botandrose that sounds like a great idea.

> `ember version` command

TIL

botandrose:
Great, I'll begin!

Yeah, I just learned about `ember version` recently myself, while struggling with unreliable builds across different machines. Its basically an attempt to bandaid the insanity that is npm and especially npm shrinkwrap. If only npm would adopt semantics like Gemfile.lock. :/

seanpdoyle:
Closed by https://github.com/thoughtbot/ember-cli-rails/commit/6eceea8a9321e3c38e2425e5f39124912ad297af.

PR #396, opened 2016-01-29
Enhance `rake ember:install` to fully reinstall if necessary.

botandrose:
Ember CLI ships with ember-cli-dependency-checker, which provides the
`ember version` command, which is akin to `bundle check`. This commit enhances
the ember:install rake task to `rm -rf node_modules bower_components` before
installing if `ember version` returns non-zero. This (helped further by npm
shrinkwrap) works to ensure that every machine has the same dependencies.

Closes [#394].

[#394]: https://github.com/thoughtbot/ember-cli-rails/issues/394

### DIFF
--- a/lib/ember_cli/path_set.rb
+++ b/lib/ember_cli/path_set.rb
@@ -78,16 +78,16 @@ module EmberCli
       end
     end
 
-    def bower_deps
-      @bower_deps ||= root.join("bower_components")
+    def bower_components
+      @bower_components ||= root.join("bower_components")
     end
 
     def npm
       @npm ||= app_options.fetch(:npm_path) { which("npm") }
     end
 
-    def npm_deps
-      @npm_deps ||= root.join("node_modules")
+    def node_modules
+      @node_modules ||= root.join("node_modules")
     end
 
     def tee
--- a/lib/ember_cli/path_set.rb
+++ b/lib/ember_cli/path_set.rb
@@ -78,8 +78,16 @@ module EmberCli
       end
     end
 
+    def bower_deps
+      @bower_deps ||= root.join("bower_components")
+    end
+
     def npm
       @npm ||= app_options.fetch(:npm_path) { which("npm") }
+    end
+
+    def npm_deps
+      @npm_deps ||= root.join("node_modules")
     end
 
     def tee
--- a/lib/ember_cli/shell.rb
+++ b/lib/ember_cli/shell.rb
@@ -54,6 +54,8 @@ module EmberCli
 
     def invalid_ember_dependencies?
       ! run("#{paths.ember} version")
+    rescue DependencyError
+      false
     end
 
     def clean_ember_dependencies!
--- a/lib/ember_cli/shell.rb
+++ b/lib/ember_cli/shell.rb
@@ -37,7 +37,8 @@ module EmberCli
         run! "#{paths.bundler} install"
       end
 
-      run! "#{paths.ember} version || rm -rf #{paths.node_modules} #{paths.bower_components}"
+      clean_ember_dependencies! if invalid_ember_dependencies?
+
       run! "#{paths.npm} prune && #{paths.npm} install"
       run! "#{paths.bower} prune && #{paths.bower} install"
     end
@@ -50,6 +51,21 @@ module EmberCli
 
     attr_accessor :pid
     attr_reader :ember, :env, :options, :paths
+
+    def invalid_ember_dependencies?
+      ! run("#{paths.ember} version")
+    end
+
+    def clean_ember_dependencies!
+      ember_dependency_directories.select(&:exist?).each(&:rmtree)
+    end
+
+    def ember_dependency_directories
+      [
+        paths.node_modules,
+        paths.bower_components,
+      ]
+    end
 
     def spawn(command)
       Kernel.spawn(
--- a/lib/ember_cli/shell.rb
+++ b/lib/ember_cli/shell.rb
@@ -37,6 +37,7 @@ module EmberCli
         exec "#{paths.bundler} install"
       end
 
+      exec "#{paths.ember} version || rm -rf #{paths.npm_deps} #{paths.bower_deps}"
       exec "#{paths.npm} prune && #{paths.npm} install"
       exec "#{paths.bower} prune && #{paths.bower} install"
     end
--- a/lib/ember_cli/shell.rb
+++ b/lib/ember_cli/shell.rb
@@ -14,7 +14,7 @@ module EmberCli
     end
 
     def compile
-      exec ember.build
+      run! ember.build
     end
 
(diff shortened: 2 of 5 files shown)


lib/ember_cli/path_set.rb
```diff
@@ -78,10 +78,18 @@ def bower
       end
     end
 
+    def bower_deps
```
seanpdoyle: What do you think about changing the names of these methods to match the directories?

`bower_deps` => `bower_components`
`npm_deps` => `node_modules`?

lib/ember_cli/shell.rb
```diff
@@ -37,6 +37,7 @@ def install
         exec "#{paths.bundler} install"
       end
 
+      exec "#{paths.ember} version || rm -rf #{paths.npm_deps} #{paths.bower_deps}"
```
seanpdoyle: While this style of invocation is compact and UNIX-y, what do you think about implementing it in Ruby, in a platform-independent manner?

I'm thinking something like:

```rb
if invalid_ember_dependencies?
  clean_dependencies!
end

# ...
private

def invalid_ember_dependencies?
  # this might return the exit status -- I'm not sure.
  # if it does, this method would check for non-zero
  ! exec("#{paths.ember} version")
end

def clean_dependencies!
  [
    paths.node_modules, 
    paths.bower_components,
  ].map(&:rmtree)
end
```

lib/ember_cli/path_set.rb
botandrose: Can do!

lib/ember_cli/shell.rb
botandrose: Hmm, looks like `exec` will raise an exception if the passed command returns non-zero. I suppose I could just use `Kernel#system`, but that doesn't provide the same guarantees or logging as `exec`. WDYT?

seanpdoyle: Although I wouldn't normally recommend it, since `exec` abstracts away so much, what do you think about catching the exception and returning `true`, otherwise returning `false`?

botandrose: The exception would be `SystemExit`... seems kinda hacky. What do you think about adding a `Runner#run` method, which does the same thing as `#run!`, but returning false instead of raising `SystemExit`?

seanpdoyle: You're absolutely right, catching a `SystemExit` is hacky.

I like your idea of extracting a `#run` method from [`#run!`](https://github.com/thoughtbot/ember-cli-rails/blob/ce5db79326f5aac72615dcb257d10e02bdc1aeb8/lib/ember_cli/runner.rb#L12-L28).

What do you think about:

```rb
def run(command)
  output, status = Open3.capture2e(@env, command, @options)

  @out.write(output)
 
  return status.success?
end

def run!(command)
  exit_successfully = run(command)

  unless exit_successfully
    @err.write <<-MSG.strip_heredoc
      ERROR: Failed command: `#{command}`
      OUTPUT:
        #{output}
    MSG
    exit 1
  end
end
```

botandrose: I like it! I think we'd also need to rename `Shell#exec` to `Shell#run!`, and introduce `Shell#run` to wrap `Runner#run`. If that sounds good to you, I'll proceed with all these changes.

seanpdoyle: so long as the methods on `Shell` remain private, I'm fine with renaming them.

botandrose:
@seanpdoyle Ready for review.

lib/ember_cli/shell.rb
```diff
@@ -34,15 +34,17 @@ def stop
 
     def install
       if paths.gemfile.exist?
-        exec "#{paths.bundler} install"
+        run! "#{paths.bundler} install"
       end
 
-      exec "#{paths.npm} prune && #{paths.npm} install"
-      exec "#{paths.bower} prune && #{paths.bower} install"
+      clean_ember_dependencies! if invalid_ember_dependencies?
```
seanpdoyle: From [thoughtbot's Ruby style guide](https://github.com/thoughtbot/guides/tree/master/style/ruby#ruby):

> Avoid conditional modifiers (lines that end with conditionals).

lib/ember_cli/shell.rb
```diff
@@ -50,6 +52,23 @@ def test
     attr_accessor :pid
     attr_reader :ember, :env, :options, :paths
 
+    def invalid_ember_dependencies?
+      !run("#{paths.ember} version")
+    rescue DependencyError
```
seanpdoyle: If a `DependencyError` was raised, what makes it more favorable to catch it and return `false` than to let it bubble and crash the program?

lib/ember_cli/shell.rb
botandrose: But, but... its so dang readable! :( I'll change it.

lib/ember_cli/shell.rb
```diff
@@ -59,13 +78,15 @@ def spawn(command)
       ) || exit(1)
     end
 
-    def exec(command)
+    delegate :run, :run!, to: :runner
```
seanpdoyle: What do you think about moving this line to the top of the `private` method declarations, near the `attr_accessor` and `attr_reader` lines?

botandrose: Sure, no problem.

lib/ember_cli/shell.rb
botandrose: This may need a comment or something else to explain the intent. This is to cover the edge case of the very first installation, before npm has installed ember itself for the first time.

lib/ember_cli/shell.rb
seanpdoyle: > But, but... its so dang readable!

We like to refer to them as `"surprise conditionals"` -- they make lines so "readable" (in an English sense) that they tend to blend in *too* well.

Conditionals (and `if` / `else` blocks) contain logic that depends on program state. They should jump out at developers.

botandrose: Ah, good points. I can understand that reasoning.

lib/ember_cli/shell.rb
botandrose: In a similar vein of `Runner#run` and `Runner#run!`, maybe the best way forward is to split `PathSet#ember` into `PathSet#ember` and `PathSet#ember!`, where the latter explodes?

botandrose:
@seanpdoyle Rebased on master to resolve conflict in CHANGELOG.md. What do you think?

seanpdoyle:
Merged in https://github.com/thoughtbot/ember-cli-rails/commit/6eceea8a9321e3c38e2425e5f39124912ad297af.

Thanks again Micah!

botandrose:
:tada:

Issue #376, opened 2016-01-03
Rails 5 API support


ivanjolic95:
Are there any plans to support Rails 5 API? The problem with it is that API controllers don't have helper_method. One solution would be to inherit gem's EmberController from a special EmberController (or whatever the name is) in app that inherits ActionController::Base instead of ActionController::API. Right now it inherits ApplicationController which is an API controller in Rails 5 API. I've configured ember-cli-rails to support testing the gem on Rails 5.0.0.beta1 locally so I can make it work that way.

ivanjolic95:
The other solution would be to add a note in readme to make ApplicationController inherit ActionController::Base when using Rails 5 API.

seanpdoyle:
@ivanjolic95 thanks for opening this issue.

I'm unfamiliar with most of the changes coming in Rails 5, so I'm having trouble following your suggestions.

Are you saying that when you `rails new --api`, your app's `ApplicationController` extends from `ActionController::API` instead of `ActionController::Base`?

ivanjolic95:
Yeah, that's right.

ivanjolic95:
And if you try to use this gem with it, there's an error because helper_method is not defined for the ActionController::API

KirillSuhodolov:
@ivanjolic95 @seanpdoyle I updated my project with rails 5 beta 1. I am using full rails(not only API) and I get 500 standard error instead of ember application and nothing in server logs.
It looks like ActionController::Base really doesn't have helper_method

ivanjolic95:
@KirillSuhodolov it's working for me when I use ActionController::Base. Which version of ember are you using?

CollinGraves:
If anyone is following this issue and needs ember-cli-rails support with a Rails 5 build, feel free to use [my fork of this repository](http://github.com/CollinGraves/ember-cli-rails) until this gets straightened out. I'll keep it as current as possible. 

Thoughts on this going forward, @seanpdoyle? Would love to send a patch through to you.

seanpdoyle:
https://github.com/thoughtbot/ember-cli-rails/pull/400 attempts to fix this issue.

seanpdoyle:
@CollinGraves @ivanjolic95 @KirillSuhodolov https://github.com/thoughtbot/ember-cli-rails/pull/400 supports Rails `5.0.0.beta2`.

Would you mind seeing if it resolves your issues?:

```rb
# Gemfile

[...]

PR #400, opened 2016-02-02
Add Rails 5 support

seanpdoyle:
Closes [#376].
	
**Improve `setup_ember` script**

* Don't overwrite `gemfiles/` changes with `appraisal install`
* Localize Ember setup to `bin/setup_ember`
* Always invoke, let `bin/setup_ember` decide whether or not to no-op

Required changes
----------------

Run test suite against the latest `5.0.0.betaX` and `master` branches of
[`rails/rails`][rails].

Since `EmberController` extends `ApplicationController`, which can now
extend from either `ActionController::Base` (which implements
`helper_method`) and `ActionController::API` (which doesn't implement
`helper_method`), we can't depend on that API existing.

Remove call to `helper_method` from `EmberController`, as it is no
longer supported.

**Fix `EmberCli::EmberController` inheritance**

The `EmberCli::EmberController` used to inherit from
`ApplicationController` in order to reap the benefits of
application-wide configurations and behavior (such as
authentication-based `before_action` calls and other macros).

Unfortunately, with the introduction of Rails 5's `rails new --api`,
applications' `ApplicationController` can now inherit from
[`ActionController::API`][api], which doesn't serve HTML by default.

To support Rails 5's default `--api` behavior,
`EmberCli::EmberController` must inherit from `ActionController::Base`,
forcing the controller to forfeit all the benefits of sharing behavior
with `ApplicationController`.

[api]: http://edgeapi.rubyonrails.org/classes/ActionController/API.html

Support changes
---------------

Rails 5 depends on `rack@2.0.x`, which **requires** Ruby `2.2.2` or
greater.

In order to support Rails 5, we will no longer support versions of Ruby
prior to `2.2.2`.

[#376]: #376
[rails]: https://github.com/rails/rails

### DIFF
--- a/bin/setup_ember
+++ b/bin/setup_ember
@@ -2,13 +2,25 @@
 
 set -e
 
-# make router catchall routes
-sed -i -e 's/auto/hash/' spec/dummy/my-app/config/environment.js
+setup_ember() {
+  local target="${1-spec/dummy/my-app}"
 
-# add text to a template
-echo 'Welcome to Ember' >> spec/dummy/my-app/app/templates/application.hbs
+  if ! [ -d $target ]; then
+    git clone -b stable https://github.com/ember-cli/ember-new-output.git $target
 
-# add an image to a template
-echo '<img src="assets/logo.png">' >> spec/dummy/my-app/app/templates/application.hbs
-mkdir -p spec/dummy/my-app/public/assets
-cp spec/fixtures/logo.png spec/dummy/my-app/public/assets
+    # make router catchall routes
+    sed -i -e 's/auto/hash/' $target/config/environment.js
+
+    # add an image to a template
+    echo '<p>Welcome to Ember</p>' >> $target/app/templates/application.hbs
+    echo '<img src="assets/logo.png">' >> $target/app/templates/application.hbs
+    mkdir -p $target/public/assets
+    cp spec/fixtures/logo.png $target/public/assets
+
+    cd $target &&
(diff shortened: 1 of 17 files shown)


bin/setup_ember
```diff
@@ -2,10 +2,24 @@
...
-sed -i -e 's/auto/hash/' spec/dummy/my-app/config/environment.js
+setup_ember() {
+  local target="${1-spec/dummy/my-app}"
 
-# add an image to a template
-echo '<img src="assets/logo.png">' >> spec/dummy/my-app/app/templates/application.hbs
-mkdir -p spec/dummy/my-app/public/assets
-cp spec/fixtures/logo.png spec/dummy/my-app/public/assets
+  if ! [ -d $target ]; then
+    git clone https://github.com/kellyselden/ember-cli-output.git $target
```
rwjblue: FWIW - This repo has recently changed, it is now at `github.com/ember-cli/ember-new-output`.

rwjblue:
Where are we at with this?  Any additional blockers?

seanpdoyle:
@rwjblue I've gotten all versions of Ruby / Rails passing except for JRuby / rails {5,master}.

I would love some help addressing [these failures](https://travis-ci.org/thoughtbot/ember-cli-rails/jobs/145094093)

seanpdoyle:
screen shot 2016-08-10 at 10 55 54 am

PR #409, opened 2016-02-26
Slight omission

kueda:


seanpdoyle:
Are you having issues running the generators?

What version of the gem are you on?

The generator namespace was renamed from `ember-cli:` to `ember:` in #344 and #345 .

kueda:
I suspect I was using the version that was in rubygems and not the version in this git repo. Sorry!

Issue #413, opened 2016-03-10
Do the heroku instructions compile app js at deploy or runtime?


nruth:
It doesn't seem to mention it and I'm not sure what the production behaviour is.

- Does it get added to the rails asset precompile
- Does it get compiled when the first user (or tens-hundreds of concurrent users) request the page after a deploy

Or could it be added to the heroku_install script after `bower install`?

seanpdoyle:
@nruth currently, the assets will be compiled at deploy time by running along with [`rake assets:precompile`](https://github.com/thoughtbot/ember-cli-rails/blob/b70fe1a6b422d782ac8068fa96164467792f5a6f/lib/tasks/ember-cli.rake#L19-L24)

nruth:
Thanks for clarifying! I'll try to add something brief to the readme.

PR #414, opened 2016-03-10
[doc] Mention when compiling happens in Heroku instructions

nruth:
based on https://github.com/thoughtbot/ember-cli-rails/issues/413

### DIFF
--- a/README.md
+++ b/README.md
@@ -172,6 +172,16 @@ task default: [:spec, "ember:test"]
 
 ## Deploy
 
+When Rails is running in production mode, EmberCLI-Rails stops doing runtime
+compilation. Instead, configured apps are built during `rake assets:precompile`.
+This keeps things quick for end users, and extends the normal Rails asset
+building process.
+
+Configuration information, including instructions for Heroku and Capistrano,
+can be found below.
+
+### CDN
+
 In production environments, assets should be served over a
 Content Delivery Network.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -270,6 +270,8 @@ You are ready to deploy:
 $ git push heroku master
 ```
 
+EmberCLI compilation happens at deploy-time, triggered by the `asset:precompilation` rake task.
```
seanpdoyle: Do you think this should be in a more prominent place?

Could you replace `EmberCLI` with `EmberCli`?

Isn't the Rake task `assets:precompile`?

nruth: I copy-pasted `EmberCLI` and `precompilation` from the Capistrano instructions to try to keep things consistent. Should I change them too?

Somewhere more prominent would be nice, especially if it's a general feature of the gem and not only something that happens on Heroku. It seems like a really nice architectural feature / benefit and worth telling people about.

seanpdoyle: the `EmberCLI` can stay for now, but the `precompilation` is incorrect and should be fixed elsewhere. Thanks!

> It seems like a really nice architectural feature / benefit and worth telling people about.
I agree! I'm open to suggestions on where to mention it.

nruth:
How about restructuring the deploy section a little? It currently opens with CDN advice, which is sound advice, but maybe not the first thing the reader is looking for. 

I'd try promoting the "Deployment Strategies" section to introduce the Deployment section, add mention of the `rake assets:precompile` -> EmberCLI build solution, 
move the CDN advice to a subsection, and keep Heroku and Capistrano as they are.

PR #421, opened 2016-03-18
Modify `Runner` to use `Open3.capture3`

seanpdoyle:
[`Open3.capture3`][docs] captures `STDOUT` and `STDERR`, allowing the
`Runner` to manually redirect their output.

This commit modified `Runner` to **only** output errors to `STDERR`
(instead of to both `STDOUT` and `STDERR`).

[docs]: http://ruby-doc.org/stdlib-1.9.3/libdoc/open3/rdoc/Open3.html#method-c-capture3

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,11 @@
 master
 ------
 
-* Remove dependency on `tee`. Fixes bug [#417][#417].
+* Only write errors to `STDERR`. [#421]
+* Remove dependency on `tee`. Fixes bug [#417][#417]. [#420]
 
+[#421]: https://github.com/thoughtbot/ember-cli-rails/issues/421
+[#420]: https://github.com/thoughtbot/ember-cli-rails/issues/420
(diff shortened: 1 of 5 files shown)


seanpdoyle:
@klaustopher does this PR resolve your noisy output issue?

Does it fail-fast?

seanpdoyle:
@thoughtbot/shell (sorry if this doesn't apply to all of you):

Can someone well-versed in Ruby pipes help me modify this implementation to use `Open3.popen3` instead of `Open3.capture3`?

Essentially, that PR “pipes” output from `STDOUT` and `STDERR` to multiple output streams (i.e. a log file and `$stdout`), with the drawback that it waits for the command to execute before writing.

Ideally, we’d stream from `STDOUT` and `STDERR` to the output streams as soon as output is available.

klaustopher:
@seanpdoyle yep, works fine ... Only outputs each error message once :+1:

PR #420, opened 2016-03-18
Remove dependency on `tee` command

seanpdoyle:
Closes [#417].

Instead of piping output to the [`tee`][tee] command (which was
previously swallowing non-zero exit statuses), capture output
from `Open3.capture2e` and write to both the specified log file and
`STDOUT`.

[#417]: https://github.com/thoughtbot/ember-cli-rails/issues/417
[tee]: http://man7.org/linux/man-pages/man1/tee.1.html

Issue #417, opened 2016-03-17
rake ember:compile returns 0 exit code even with failed compilation


wireframe:
compiling the app via rake appears like everything is fine, but grepping my `log/ember-appname.development.log` log file is littered with error messages like this:

```
Package: json-stable-stringify
Required by: frontend / broccoli-asset-rev
  * Specified: 1.0.1
  * Installed: (not installed)
```

```bash
$ rake ember:compile && echo $?
$ 0
```

raising an error with a truncated (or full) error from the shell/log would be *hugely* helpful.

seanpdoyle:
@wireframe thanks, this is certainly a bug.

I'll look into it.

seanpdoyle:
@wireframe on further inspection, things might be behaving properly.

The [`ember:compile` Rake task depends on `ember:install`](https://github.com/thoughtbot/ember-cli-rails/blob/b70fe1a6b422d782ac8068fa96164467792f5a6f/lib/tasks/ember-cli.rake#L3), which would install the missing dependencies. 

Other than the logs containing output about missing dependencies (which might come from [`ember-cli-dependency-checker`](https://github.com/quaertym/ember-cli-dependency-checker/blob/6a3d4d11da15921e95376e2d2431fa9ed23b971e/lib/reporter.js#L34-L35)), what behavior are you seeing that indicates that the build is failing?

wireframe:
after running `rake ember:compile` with the failed state, I reload my rails app and get an error from the ember controller saying that the index.html file can not be found.

seanpdoyle:
@wireframe I've tried to reproduce your issues with https://github.com/seanpdoyle/ember-cli-rails-heroku-example, but I'm having some trouble. The `ember:install` Rake dependency seems to be preventing me from compiling without a necessary dependency.

Could you pull https://github.com/seanpdoyle/ember-cli-rails-heroku-example down and try to reproduce? If you're able to, could you share the steps it took on this issue?

klaustopher:
I just ran into an issue as well ... I had an issue on one machine, that `~/.config/configstore/ember-cli.json` was owned by root and ember-cli could not change it.

```
$ RAILS_ENV=staging bin/rake ember:compile
$ echo $?
0
```

The command finishes successfully, but the app is not available and if I look into the log file, this:

```
$ cat log/ember-demandcheck.staging.log
[...]

PR #423, opened 2016-03-18
Redirect output across streams

seanpdoyle:
`Runner` will "stream" output to `STDOUT`, `STDERR`, and each of the
associated streams instead of waiting until the subprocesses complete.

### DIFF
--- a/lib/ember_cli/runner.rb
+++ b/lib/ember_cli/runner.rb
@@ -4,18 +4,22 @@ module EmberCli
   class Runner
     def initialize(out:, err:, env: {}, options: {})
       @env = env
-      @out_streams = Array(out)
-      @err_streams = Array(err)
+      @output_streams = Array(out)
+      @error_streams = Array(err)
       @options = options
+      @threads = []
     end
 
     def run(command)
-      output, error, status = Open3.capture3(env, command, options)
+      Open3.popen3(env, command, options) do |stdin, stdout, stderr, process|
+        stdin.close
 
-      write(output, streams: out_streams)
-      write(error, streams: err_streams)
+        threads << redirect_stream_in_thread(stdout, write_to: output_streams)
+        threads << redirect_stream_in_thread(stderr, write_to: error_streams)
 
-      status
+        threads.each(&:join)
+        process.value
+      end
     end
 
     def run!(command)
@@ -28,13 +32,17 @@ module EmberCli
 
     protected
 
-    attr_reader :env, :err_streams, :options, :out_streams
+    attr_reader :env, :error_streams, :options, :output_streams, :threads
 
     private
 
-    def write(output, streams:)
-      streams.each do |stream|
-        stream.write(output)
+    def redirect_stream_in_thread(stream, write_to:)
+      Thread.new do
+        Thread.current.abort_on_exception = true
+
+        while line = stream.gets
+          write_to.each { |redirection_stream| redirection_stream.puts(line) }
+        end
       end
     end
   end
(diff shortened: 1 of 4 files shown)


lib/ember_cli/runner.rb
```diff
@@ -32,9 +35,11 @@ def run!(command)
 
     private
 
-    def write(output, streams:)
-      streams.each do |stream|
-        stream.write(output)
+    def redirect(stream, pipes:)
+      Tee.open(*pipes, stdout: nil) do |pipe|
+        while line = stream.gets
```
seanpdoyle: Is this the best way to redirect `STDERR` / `STDOUT` to multiple output streams? The use case is to redirect `STDOUT` to both a log file and `STDOUT` (which explains the dependence on a `tee` Ruby implementation, given [the trouble we've had with using a shell `tee` implementation](https://github.com/thoughtbot/ember-cli-rails/issues/417#issuecomment-197991094))

@thoughtbot/shell I'm unsure if this `while` loop (within a Ruby block) will block other output streams (imagine a case where a command is writing to `STDOUT`, encounters an error, wants to write to `STDERR`, then continue to write to `STDOUT`)

bernerdschaefer: I don't know the `tee` gem or too much about the nuances Ruby's process-running features, so it's hard to weigh in on the general approach.

It's not **generally** a problem with blocking, as `STDOUT` and `STDERR` have pretty large buffers which protect writers from being blocked. But it is possible.

This will also stream all `STDOUT` messages first, then all `STDERR` messages, which may be unexpected.

seanpdoyle: @bernerdschaefer this PR is a follow up to https://github.com/thoughtbot/ember-cli-rails/pull/420.

The idea here is that streaming up-to-date information is better than waiting until the subprocess completes, then dumping all output.

Do you have suggestions for a better approach [that also handles `STDOUT` / `STDERR` redirection from Ruby-land](https://github.com/thoughtbot/ember-cli-rails/issues/417#issuecomment-197991094)?

jferris: In capybara-webkit, we used [IO.copy_stream](http://ruby-doc.org/core-2.3.0/IO.html#method-c-copy_stream) in a background thread.

https://github.com/thoughtbot/capybara-webkit/blob/master/lib/capybara/webkit/connection.rb#L103

seanpdoyle: @jferris thanks for the suggestion.

I've tried `copy_stream` before, but not from a background thread -- is the Thread necessary?

jferris: I believe `copy_stream` is blocking, so it won't ever return for an infinite stream.

lib/ember_cli/runner.rb
```diff
@@ -32,9 +34,13 @@ def run!(command)
...
-    def write(output, streams:)
-      streams.each do |stream|
-        stream.write(output)
+    def redirect(stream, pipes:)
+      pipes.each do |pipe|
+        Thread.new do
+          Thread.current.abort_on_exception = true
+
+          IO.copy_stream(stream, pipe)
+        end
```
seanpdoyle: @jferris thanks for the tip about `IO.copy_stream`.

Does this implementation look ok?

seanpdoyle: @jferris on top of that, now that we've introduce asynchrony, what's a good way to test this?

jferris: If you close the input stream, `copy_stream` will end. If you want to wait until the thread is finished, you'll have to store the thread somewhere and call `wait`.

PR #439, opened 2016-04-18
Inquire about application server in issue template

seanpdoyle:
`ember-cli-rails` [has issues with multi-process or multi-thread
application servers][commit], so having that information available at
the time of issue is very valuable.

[commit]: https://github.com/thoughtbot/ember-cli-rails/commit/ef0a49546a8c2e5fb0a298ba455fba97289065a1

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -12,6 +12,11 @@
 
 **What is the [`ember-cli-rails-addon`][addon] version (from `package.json`)?**
 
+**What is your application server (e.g. `puma`, `unicorn`, `thin`, `webrick`,
+etc.)?**
+
+**Is your application server multi-threaded or multi-process?**
+
 **What are the contents of `config/initializers/ember.rb`?**
 
 **What are the contents of the Rails' view that renders the Ember application?**


.github/ISSUE_TEMPLATE.md
```diff
@@ -12,6 +12,11 @@
 
 **What is the [`ember-cli-rails-addon`][addon] version (from `package.json`)?**
 
+**What is your application server (e.g. `puma`, `unicorn`, `thin`, `webrick`,
+etc.)?**
+
+**Is your application server multi-threaded or multi-process?**
```
mike-burns: This phrasing makes me think that I have to pick exactly one of these two options. Is that correct? Do we need to know this separately from the above question?

seanpdoyle: @mike-burns good questions.

Answers to this question would provide the most useful information. The previous question is mostly a setup question, with this one as the follow up.

I split them into two because I feared that this question on its own would yield answers like "I don't know", since some developers might not put a lot of thought into choosing (or thinking about) their application server.

Would combining these two questions be an improvement?

Would omitting the list of examples be an improvment?

Could this question effectively stand on its own?

mike-burns: How about one question, perhaps something like:

> Is your application server multi-threaded (such as `puma` and `unicorn`) or is it multi-process (such as `thin` and `webrick`)?

(It could be that I still don't understand if this is the right question to ask.)

seanpdoyle: Maybe.

To give you more background, `ember-cli-rails` will manage an `ember server` subcommand when run in development mode. 

Running on a multi-{process,thread} server, the gem would run an `ember server` subcommand for each worker {process,thread}, which would end up clobbering each other. This leads to a 1 in N (where N is the number of workers) chance that your request for EmberCLI-generated assets will be successful.

> (It could be that I still don't understand if this is the right question to ask.)

Did that explanation help? Does the question in the template feel like the right thing to ask?

mike-burns: Great explanation. Does it appear elsewhere? Can one of these questions link to that explanation?

seanpdoyle:
Merged in https://github.com/thoughtbot/ember-cli-rails/commit/a97145d1ed372a7cd7aad3e44f4c613cb350de69

PR #449, opened 2016-05-12
Ignore requests for /rails/info routes

stevenharman:
When mounting an Ember CLI app via `mount_ember_app`, the catch-all routes defined by Ember CLI Rails also catch requests to two of Rails' magic routes:

1. `/rails/info/routes` (which is also redirected to from `/rails/info`)
2. `/rails/info/properties`

This adds another constraint to the mounted Ember app that will ignore requests with a path starting with `/rails/info`. I'm not in love with this solution, and basically hacked it together to see if it's even functionality you'd like added to the gem. If so, we can work to clean up the implementation, add documentation, etc... If not, perhaps we can simply document this in the `README` and maybe include the constraint so others can apply it if they wish.

Let me know what I can do to help.

### DIFF
--- a/lib/ember_cli/route_helpers.rb
+++ b/lib/ember_cli/route_helpers.rb
@@ -1,14 +1,10 @@
 require "ember_cli/html_constraint"
+require "ember_cli/multi_constraint"
 require "ember_cli/rails_info_constraint"
 
 module ActionDispatch
   module Routing
     class Mapper
-      MULTI_CONSTRAINT = -> (req) {
-        [::EmberCli::HtmlConstraint.new,
-         ::EmberCli::RailsInfoConstraint.new].all? { |c| c.matches?(req) }
-      }
-
       def mount_ember_app(app_name, to:, **options)
         routing_options = options.deep_merge(
           defaults: { ember_app: app_name },
@@ -20,7 +16,8 @@ module ActionDispatch
           format: :html,
         )
 
-        scope constraints: MULTI_CONSTRAINT do
+        scope constraints: ::EmberCli::MultiConstraint.new(::EmberCli::HtmlConstraint.new,
+                                                           ::EmberCli::RailsInfoConstraint.new) do
           get("#{to}(*rest)", routing_options)
         end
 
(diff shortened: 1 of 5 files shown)


lib/ember_cli/route_helpers.rb
```diff
@@ -1,8 +1,14 @@
 require "ember_cli/html_constraint"
+require "ember_cli/rails_info_constraint"
 
 module ActionDispatch
   module Routing
     class Mapper
+      MULTI_CONSTRAINT = -> (req) {
```
stevenharman: Really? I'm personally a fan of the idea @jimweirich put forward about blocks (paraphrased):

> Curly brackets are for functional blocks, where the primary purpose of the block is to return a value. `do ... end` is for procedural blocks, where the primary purpose of the block is to have a side-effect.

That said, I'm happy to comply with your style guide if it pleases the bots.

seanpdoyle:
@stevenharman  sorry that this PR has sat for so long.

I think `master` is green once again, would you mind rebasing and pushing to get CI passing?

stevenharman:
@seanpdoyle I've rebased the PR. Are there any changes you'd like me to make? I'm sure the naming could use some improvement. I didn't put too much effort into the initial PR as I wasn't sure this was something you'd want in the project.

I'm happs to make any changes you see fit.

lib/ember_cli/route_helpers.rb
seanpdoyle: What do you think about extracting a class instead of using a lambda?

lib/ember_cli/route_helpers.rb
```diff
@@ -14,7 +16,8 @@ def mount_ember_app(app_name, to:, **options)
           format: :html,
         )
 
-        scope constraints: ::EmberCli::HtmlConstraint.new do
+        scope constraints: ::EmberCli::MultiConstraint.new(::EmberCli::HtmlConstraint.new,
```
stevenharman: I'm not sure what to do about this one as I can't find an aesthetically pleasing way to format this one. Suggestions?

seanpdoyle:
Thanks for opening this PR and submitting this code.

You've inspired me to try and address the problem myself in a separate PR.

Could you provide some feedback on https://github.com/thoughtbot/ember-cli-rails/pull/471?

seanpdoyle:
Closing in favor of https://github.com/thoughtbot/ember-cli-rails/pull/471#issuecomment-231433077.

Thanks @stevenharman!

Issue #451, opened 2016-05-17
Possible memory issue


egh:
**Description**

While debugging some memory issues in our application, I came across a possible memory issue in ember-cli-rails. It seems that ember-cli-rails is recompiling the ember template every time a request is made, rather than reusing the cached the erb template. This is a problem both because it is slow and because rails caches the compiled template. Re-compiling every time will add to the memory of the usage of the application over time.

This can be confirmed either by adding a breakpoint to line 295 of `actionview-4.2.6/lib/action_view/template.rb` or by checking the memory usage using derailed.

This is a script to reproduce the issue in a brand new ember-cli-rails app:

```
#/bin/sh -xe
rails new ember_cli
cd ember_cli
echo 'gem "ember-cli-rails"' >> Gemfile
bundle install
ember new frontend --skip-git
rails generate ember:init
cd frontend
ember install ember-cli-rails-addon
cd ..
cat > config/database.yml <<EOF
production:
  adapter: sqlite3
  database: db/prod.sqlite3
  pool: 5
  timeout: 5000
EOF
cat > config/routes.rb<<EOF
Rails.application.routes.draw do
  mount_ember_app :frontend, to: "/"
end
EOF
cat > config/secrets.yml<<EOF
production:
  secret_key_base: 39fbddcabac94a6365f87044451c9bbb3c29b5fe8c70c27edada5ca624e5dbb9ab247f8b811ea2e20244c7d084a8c3d5d86f40ae6a198a0df28700ec4fba32a5
EOF

rake ember:install
rake ember:compile

# benchmark

echo 'gem "derailed"' >> Gemfile
bundle install
```

After generating the app, if can be tested using:
`TEST_COUNT=1000 RAILS_ENV=production bundle exec derailed exec perf:objects`

This will the root of the app 1000 times and report back on the memory usage. Of particular interest is the line:

> Total retained:  3411369 bytes (9877 objects)

and:

> retained memory by location
> -----------------------------------
>    2928842  /home/egh/.rbenv/versions/2.2.4/lib/ruby/gems/2.2.0/gems/actionview-4.2.6/lib/action_view/template.rb:296
>     240000  /home/egh/.rbenv/versions/2.2.4/lib/ruby/gems/2.2.0/gems/actionview-4.2.6/lib/action_view/template.rb:107
[...]

PR #452, opened 2016-05-19
Use render html:, not render inline:

egh:
Render inline compiles a template and uses more memory.
It also takes slightly more time.

See: https://github.com/thoughtbot/ember-cli-rails/issues/451

### DIFF
--- a/app/helpers/ember_rails_helper.rb
+++ b/app/helpers/ember_rails_helper.rb
@@ -8,6 +8,6 @@ module EmberRailsHelper
 
     head, body = markup_capturer.capture
 
-    render inline: EmberCli[name].index_html(head: head, body: body)
+    render html: EmberCli[name].index_html(head: head, body: body).html_safe
   end
 end


egh:
A few rough benchmarks. I ran 1k requests against a webrick served app:

before this change, 1k requests took 2m5.936s, and the RSS was 104384.

after this change,  1k requests took 2m1.679s, and the RSS was 97120

seanpdoyle:
Thanks for investigating #451!

Will rendering as `text` still be interpreted as HTML by the consuming browser?

egh:
I believe `:text` is supposed to inherit whatever the parent does. However, in this case it seems `:html` would be more appropriate. I will change and force push.

Those test failures appear to be present on master as well.

seanpdoyle:
@egh sorry that this PR has sat for so long.

I think `master` is green once again, would you mind rebasing and pushing to get CI passing?

seanpdoyle:
Merged in https://github.com/thoughtbot/ember-cli-rails/commit/a4bddef2a3c51dd603b1715adf7bfba8e8e89625.

Thanks!

egh:
Thank you!

PR #455, opened 2016-05-26
Fix dependencies check for compile command

krasnoukhov:
First commit fixes actual bug, while the second one improves `clean_ember_dependencies!` behavior when dependencies directory is a symlink.

We use shared directories for `node_modules` and `bower_components` in order to speed up deploys. This change make cleanup work for that case.

Thanks!

### DIFF
--- a/lib/ember_cli/shell.rb
+++ b/lib/ember_cli/shell.rb
@@ -63,14 +63,14 @@ module EmberCli
     end
 
     def clean_ember_dependencies!
-      ember_dependency_directories.select(&:exist?).each(&:rmtree)
+      ember_dependency_directories.flat_map(&:children).each(&:rmtree)
     end
 
     def ember_dependency_directories
       [
         paths.node_modules,
         paths.bower_components,
-      ]
+      ].select(&:exist?)
     end
 
     def spawn(command)
--- a/lib/ember_cli/shell.rb
+++ b/lib/ember_cli/shell.rb
@@ -57,7 +57,7 @@ module EmberCli
     delegate :run, :run!, to: :runner
 
     def invalid_ember_dependencies?
-      !run("#{paths.ember} version")
+      !run("#{paths.ember} version").success?
     rescue DependencyError
       false
     end


lib/ember_cli/shell.rb
```diff
@@ -57,13 +57,14 @@ def test
...
-      !run("#{paths.ember} version")
+      !run("#{paths.ember} version").success?
     rescue DependencyError
       false
     end
 
     def clean_ember_dependencies!
-      ember_dependency_directories.select(&:exist?).each(&:rmtree)
+      dirs = ember_dependency_directories.select(&:exist?).flat_map(&:children)
+      dirs.each(&:rmtree)
```
seanpdoyle: Thanks for opening this PR @krasnoukhov !

I apologize for letting this method get a little out of hand.

What do you think about extracting some private helper methods to more clearly communicate this method's intent?

Would this be a good start?:

```ruby
def ember_dependency_directories
  [
    paths.node_modules,
    paths.bower_components,
  ].select(&:exist?)
end

def clean_ember_dependencies!
  ember_dependency_directories.flat_map(&:children).each(&:rmtree)
end
```

krasnoukhov: That makes total sense. I've updated the commit

krasnoukhov:
It does not seem specs fail because of my change. @seanpdoyle is there anything else I could do in order to get this merged?

seanpdoyle:
@krasnoukhov unfortunately, it looks like `master` is broken as well.

I'll try to dedicate some time soon to get `master` passing so that development can continue.

I'll keep you posted!

krasnoukhov:
@seanpdoyle I went ahead and added a PR for specs fixing: #456

krasnoukhov:
@seanpdoyle This is green now. Can you cut a bugfix release once you merge it? Thanks

seanpdoyle:
Merged in https://github.com/thoughtbot/ember-cli-rails/releases/tag/v0.7.4.

Released in [`0.7.4`](https://github.com/thoughtbot/ember-cli-rails/releases/tag/v0.7.4).

Thanks!

krasnoukhov:
Thanks :green_heart:

PR #456, opened 2016-05-26
Make specs green again!

krasnoukhov:
Seems like `ember-cli-output` app does not include default
text in `applicition.hbs` template, so we add it anyway.


krasnoukhov:
@seanpdoyle This one should be ready

seanpdoyle:
PR 4-5-6 indeed.

krasnoukhov:
😆

PR #459, opened 2016-06-06
adding deployment performance tip for Capistrano deploy

dchersey:
I've noticed that capistrano deployments rebuild bower and node modules from scratch each time, so I experimented with sharing those directories across deployments.  Seems to be working, so I thought I would update the Capistrano section with this tip for others.

### DIFF
--- a/README.md
+++ b/README.md
@@ -303,6 +303,13 @@ you're not using `nvm`, make sure the string you prepend to the `$PATH` variable
 contains the directory or directories that contain the `bower` and `npm`
 executables.
 
+#### For faster deployments
+Place the following in your deploy/<environment>.rb 
+```ruby
+set :linked_dirs, %w{frontend/node_modules frontend/bower_components}
+```
+to avoid rebuilding all the node modules and bower components with every deploy
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -303,6 +303,13 @@ you're not using `nvm`, make sure the string you prepend to the `$PATH` variable
 contains the directory or directories that contain the `bower` and `npm`
 executables.
 
+#### For faster deployments
+Place the following in your deploy/<environment>.rb 
+```ruby
+set :linked_dirs, %w{frontend/node_modules frontend/bower_components}
```
seanpdoyle: @dchersey thanks for adding this information!

`frontend` is the default name of the Ember application, but it's an arbitrary value that can be configured.

Would you mind updating this text to account for that?

dchersey:
Done!

seanpdoyle:
Merged in https://github.com/thoughtbot/ember-cli-rails/commit/96a3aa6f4e158bd8c8c4d6daf982cc469a85a3ef.

Thanks!

Issue #464, opened 2016-06-21
Ember cli rails breaks when deprecations(warnings) are found on Ember


victor95pc:
Linux 64-bits, Manjaro(Arch Linux)
ruby 2.1.8
npm 3.9.5
ember-cli 2.6.2
rails 4.2.6
ember-cli-rails gem 0.7.4
ember-cli-rails-addon 0.7.0
application server webrick

```
#config/initializers/ember.rb
EmberCli.configure do |c|
  c.app :frontend
end
```

contents of the Rails' view
```
<%= render_ember_app :frontend do |head, body| %>
<% end %>
```
 route.rb
  ```
mount_ember_app    :frontend, to: '/cursos', controller: 'ember_loader'
mount_ember_assets :frontend, to: '/'
```

There is a problem on ember-cli-rails, every deprecation raise like a BuildError, which is not good, a deprecation example: 
`DEPRECATION: Overriding init without calling this._super is deprecated`

victor95pc:
Pull Request: #465

seanpdoyle:
@victor95pc https://github.com/thoughtbot/ember-cli-rails/pull/468 should address this.

Would you mind pulling down `master` and testing whether it has been resolved?

victor95pc:
@seanpdoyle @mhaylock Thx guys for your work, sorry for the delay, this PR work perfect and cover all use cases.

seanpdoyle:
@victor95pc thanks for clarifying!

PR #465, opened 2016-06-21
Ignore deprecation(warnings) on Ember

victor95pc:
#464


lib/ember_cli/build_monitor.rb
```diff
@@ -36,7 +36,11 @@ def complete?
     end
 
     def build_error?
-      error_file.exist? && error_file.size?
+      error_file.exist? && error_file.size? && not(has_deprecation_warnings?)
```
mhaylock: If the error file contains a deprecation warning (and it always will in the case of the deprecation warning from #464) then this `not(has_deprecation_warnings?)` check will mean that even if the error file contains actual build errors as well `build_error?` will still respond `false`.

Instead the check should verify whether the error file contains any lines that are not a deprecation warning.

mhaylock: Also look at `EmberCli::BuildMonitor#reset!` - this method is used to empty the error file between builds. Internally it uses `build_error?`, but even when a deprecation warning is seen it will still need to be deleted. The method needs to be refactored to delete the file regardless of whether the contents are deprecation warnings or build errors.

victor95pc: thanks for your explanation @mhaylock

seanpdoyle:
Closing this in favor of https://github.com/thoughtbot/ember-cli-rails/pull/468 (which was merged into `master` in https://github.com/thoughtbot/ember-cli-rails/commit/448bae91ee0608baa22204294bbc4eeec05a6928)

PR #466, opened 2016-06-28
Include helper & extra modules when using API.

undergroundwebdesigns:
Adds logic to conditionally load extra controller modules and manually include the EmberCLI helper when inheriting from an ApplicationController that inherits from ActionController::API instead of ActionController::Base. Allows the gem to work as expected with Rails 5 apps that use the `--api` option when generating the app.


williamweckl:
+1

seanpdoyle:
Closing in favor of #400.

If you think this shouldn't be closed, please comment and I'll reopen.

PR #468, opened 2016-07-02
Add support for detecting DEPRECATION: warnings

### LINKED ISSUE #464: Ember cli rails breaks when deprecations(warnings) are found on Ember

mhaylock:
These warnings aren't build errors (the build is still successful), so we need to avoid treating such `DEPRECATION:` lines as indicating a build error occurred.

This fixes #464 - a problem I've encountered with a new project using ember-cli-rails 0.7.4, Ember CLI 2.6.2 and ember-cli-rails-addon 0.7.0 (which is the source of the deprecation warning).

A different solution has been presented in PR #465 - this PR differs in that it will still detect build errors that occur even when a deprecation warning is present. It also ensures the build error files is still cleared between builds.

Thanks to @victor95pc for your work on #465 - it helped steer me in the right direction for fixing this :)

For the curious (because this is something I wondered myself) - the STDERR output of `ember build` is used to detect a build error instead of the exit status because in development mode the build is run with `--watch` and it never exits for there to be an exit status to assess.

### DIFF
--- a/lib/ember_cli/build_monitor.rb
+++ b/lib/ember_cli/build_monitor.rb
@@ -14,7 +14,7 @@ module EmberCli
     end
 
     def reset
-      if build_error?
+      if build_error_output?
         error_file.delete
       end
     end
@@ -35,8 +35,21 @@ module EmberCli
       !paths.lockfile.exist?
     end
 
-    def build_error?
+    def build_error_output?
       error_file.exist? && error_file.size?
+    end
+
+    def build_error?
+      return false unless build_error_output?
+      lines = error_file.readlines
+      # Filter out blank and backtrace lines (leaving just messages):
+      lines.reject! { |line| line =~ /(^\s*$|^\s{4}at .+$)/ }
+      # Filter out any deprecation warnings (allowing for ASCII color escape
+      # codes at the start of the line):
+      lines.reject! { |line| line =~ /^(\e[^\s]+)?DEPRECATION:/ }
+
+      # There was a build error if any lines remain:
+      lines.any?
     end
 
     def error_file
--- a/spec/lib/ember_cli/build_monitor_spec.rb
+++ b/spec/lib/ember_cli/build_monitor_spec.rb
@@ -15,6 +15,18 @@ describe EmberCli::BuildMonitor do
       end
     end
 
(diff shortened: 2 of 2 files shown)


mhaylock:
The CI failures on this PR are due to the fact that those specs are also broken on master - the specs are still testing Rails 3.2 but the code does not support it. I've created a separate PR to fix this: #469.

victor95pc:
Thx @mhaylock, my solution was not good enough? Can you explain a little more?

mhaylock:
Hi @victor95pc, thanks again for your solution. There are couple of problems it doesn't address, but would be simple enough to fix. Apologies if I've cause any offence by presenting my own solution.

As you've requested further explanation I'll comment on your PR to clarify where I see there being some fixes required.

victor95pc:
No problem at all, I take about 30 minutes to create that solution, I didn't thought too much about all use cases, I just thought my solution was more clean but yours cover better all use cases, thanks for giving your time on this awesome project I love. .

mhaylock:
Thanks to @seanpdoyle merging #469 I've been able to rebase this PR and now the specs are passing.

seanpdoyle:
@mhaylock I've cleaned up and squashed your commits into https://github.com/thoughtbot/ember-cli-rails/commit/448bae91ee0608baa22204294bbc4eeec05a6928.

Thanks for this PR!

mhaylock:
Excellent, thanks @seanpdoyle!

PR #469, opened 2016-07-02
Remove Rails 3.2 from Specs

mhaylock:
Rails 3.2 is no longer supported [according to the README](https://github.com/thoughtbot/ember-cli-rails/blob/a4bddef2a3c51dd603b1715adf7bfba8e8e89625/README.md#ruby-and-rails-support), and the specs fail because EmberRailsHelper now uses ActionView::Template::HTML (not available before Rails 4.1) - a change introduced in a4bddef2a3c51dd603b1715adf7bfba8e8e89625.

This PR removes the specs for Rails 3.2.

If the README is inaccurate and Rails 3.2 should be supported then to fix it `EmberRailsHelper#render_ember_app` would need to be updated so that:
```ruby
render html: EmberCli[name].index_html(head: head, body: body).html_safe
```
becomes something like:
```ruby
if defined?(ActionView::Template::HTML)
  render html: EmberCli[name].index_html(head: head, body: body).html_safe
else
  render inline: EmberCli[name].index_html(head: head, body: body)
end
```


seanpdoyle:
This looks great!

Thanks!

PR #471, opened 2016-07-08
Don't route `/rails/info` request to Ember

seanpdoyle:
Closes [#449].

> When mounting an Ember CLI app via `mount_ember_app`, the catch-all
> routes defined by Ember CLI Rails also catch requests to two of Rails'
> magic routes:
>
> 1. `/rails/info/routes` (which is also redirected to from `/rails/info`)
> 2. `/rails/info/properties`
>
> This adds another constraint to the mounted Ember app that will ignore
> requests with a path starting with `/rails/info`.
>
> @stevenharman

[#449]: https://github.com/thoughtbot/ember-cli-rails/pull/449


stevenharman:
Ah... I was wondering if this was what you were asking for when you mentioned moving from a lambda to a discrete object. LGTM.

PR #492, opened 2016-10-06
doc: heroku slug pruning with buildpack

### LINKED ISSUE #491: Heroku slug size: can node_modules be deleted after build?
3 apps takes the slug size up towards 300MB, and it looks like most of this is in node_modules that shouldn't be necessary after the precompile step creates the javascript and css assets?

Poking around with `heroku run bash` and `du` is showing the node_modules directories as being around 200-500MB which will zip to some extent but shouldn't be necessary.

Is there a reason I've not thought of for keeping these around in the final dyno slug, or could we add something to delete the files from the slug once asset compilation succeeds?

There's something similar in the [Heroku docs for Java](https://devcenter.heroku.com/articles/reducing-the-slug-size-of-java-applications).

nruth:
Hi Sean

Please take a look and see if this text is ok.

I linked to the issue rather than writing up a summary. This helps keep it short but might not be what you want.

Any feedback welcome.

closes #491

### DIFF
--- a/README.md
+++ b/README.md
@@ -278,6 +278,11 @@ applications into the project.
 
 [buildpack]: https://devcenter.heroku.com/articles/using-multiple-buildpacks-for-an-app#adding-a-buildpack
 
+#### Slug size
+
+[Heroku slug size is limited](https://devcenter.heroku.com/articles/slug-compiler#slug-size). We leave behind some assets after the Rails precompile task, and these are included in the Heroku slug.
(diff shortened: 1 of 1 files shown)


seanpdoyle:
Thanks for doing the leg work for this!

Merged in https://github.com/thoughtbot/ember-cli-rails/commit/631a161213156d43c707e71e7a722f25cc3a7484.

PR #498, opened 2016-10-23
baseURL deprecated in favor of rootURL

JCBarry:
Ran into this when configuring my ember-cli-rails Ember application under something other than `/`.  There is some history to this deprecation that can be found here and it's only relevant in ember-cli versions > 2.7. 

Check out this relevant official blog post: http://emberjs.com/blog/2016/04/28/baseURL.html

### DIFF
--- a/README.md
+++ b/README.md
@@ -466,7 +466,7 @@ Rails.application.routes.draw do
 end
 ```
 
-Then set each Ember application's `baseURL` to the mount point:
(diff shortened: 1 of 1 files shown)


seanpdoyle:
Merged in https://github.com/thoughtbot/ember-cli-rails/commit/c956653bca6a379e1e0673e90854ad5ed067ebe5.

Thanks!

PR #502, opened 2016-11-08
Improve EmberController inheritance change wording

jswanner:
The previous wording implies that application-wide behavior would no longer apply to `EmberCli::EmberController`, due to inheritance change. But, `EmberCli::EmberController` still inherits from `ApplicationController`, it just no longer assumes that `ApplicationController` inherits from `ActionController::Base`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,8 +25,8 @@ master
 0.8.0
 -----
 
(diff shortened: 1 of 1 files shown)


seanpdoyle:
Thanks!

Issue #503, opened 2016-11-25
mount_ember_app with custom path - No route matches [GET] "/assets/frontend.js"


ndbroadbent:
**Which operating system and version is the project developed on?**

Mac OS X 10.11.6 (15G1108)

**Which version of [`ruby`][ruby] is the project developed on?**

2.2.3p173

**Which version of [`npm`][npm] is the project developed on?**

4.0.2

**Which version of [`ember-cli`][ember-cli] is the project developed on?**

ember-cli: 2.9.1
node: 4.4.4
os: darwin x64

**What is the [`rails`][rails] version?**

5.0.0.1

**What is the [`ember-cli-rails`][gem] version (from `Gemfile`)?**

0.8.3

**What is the [`ember-cli-rails-addon`][addon] version (from `package.json`)?**

0.8.0

**Is your application server [multi-threaded][] (such as `puma` and `unicorn`) or is
it multi-process (such as thin and webrick)?**

puma (3.6.2)

**What are the contents of `config/initializers/ember.rb`?**

```ruby
EmberCli.configure do |c|
  c.app :frontend
end
```

**What are the contents of the Rails' view that renders the Ember application?**

I am just rendering the default index from `EmberCli::EmberController`:

```html
<html>
  <head>
    <meta charset="utf-8">
    <meta http-equiv="X-UA-Compatible" content="IE=edge">
    <title>Frontend</title>
    <meta name="description" content="">
    <meta name="viewport" content="width=device-width, initial-scale=1">

    

    <link rel="stylesheet" href="/assets/vendor.css">
    <link rel="stylesheet" href="/assets/frontend.css">

    
      <meta name="csrf-param" content="authenticity_token" />
<meta name="csrf-token" content="msN73mx+hlq9gXKywClNUIAe1xk1vkVxPYBlIqTsZIyti9MLAHojI/WtDi8FmPWpPxLJi87vK1O/MsgC5U2BYA==" />
</head>
  <body>
    

    <script src="/assets/vendor.js"></script>
    <script src="/assets/frontend.js"></script>

    
  </body>
</html>
```

**How are the EmberCLI-related routes defined?**

`config/routes.rb`:

```ruby
Rails.application.routes.draw do
  root to: 'marketing#index'
  devise_for :users
  ActiveAdmin.routes(self)

  mount_ember_app :frontend, to: "/app"  
end
```

**How is the application deployed?**

Developing locally

---------------------------------------------------

[...]

PR #504, opened 2016-11-25
Add note about changing rootURL when using a custom path

ndbroadbent:
I'm new to ember-cli-rails, and this is the first issue I ran into (see #503)

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,6 +132,7 @@ end
   `format: :html` requests. Additionally, this will handle child routes as well.
   For instance, mounting `mount_ember_app :frontend, to: "/frontend"` will handle a
   `format: :html` request to `/frontend/posts`.
+  *Note:* If you specify a custom path, you must also update the `rootURL` in `frontend/config/environment.js`. See [Mounting multiple Ember applications](#mounting-multiple-ember-applications) for more information.
 * `controller` - Defaults to `"ember_cli/ember"`
 * `action` - Defaults to `"index"`
 


dirtyhenry:
I ran into this issue. The rootURL in Ember's config file needs a trailing `/` character or you'll get a warning from Ember. In the example from the README file, the trailing `/` is missing. Maybe it's worth updating as well.

ndbroadbent:
Oh yeah I ran into the trailing slash issue too. I was pretty confused about that. I just don't understand why it's necessary. It seems like a very easy fix, but the Ember developers decided that a trailing slash is very important? What's the reasoning?

It's also pretty annoying, because you can't force a trailing slash in Rails routes. Rails strips all trailing slashes before it hits your routes, so there's no way to check if it is present.

seanpdoyle:
Thanks for opening this PR, @ndbroadbent!

Would you mind addressing https://github.com/thoughtbot/ember-cli-rails/pull/504#issuecomment-263247745 in this PR as well?

nruth:
We use [rack-rewrite](https://github.com/jtrupiano/rack-rewrite) to handle the trailing / issue. 

It's bemusing to strip / off for rails pages but keep them for ember pages, but seems to work:

```
    #remove trailing slashes
    r301 %r{\A/(?!emberapp1)(?!emberapp2)(.*)/\Z}, '/$1'

    # otherwise ember crashes with root-url not containing /
    r302 '/emberapp1', '/emberapp1/'
    r302 '/emberapp2', '/emberapp2/'

```

PR #519, opened 2017-03-12
Patch Rails 5.1 not supporting text option for render anymore

weggesser:
Ran into an exception while trying to use this with rails 5.1 - :text option is no longer supported by ActionView:: TemplateRenderer

### DIFF
--- a/app/helpers/ember_rails_helper.rb
+++ b/app/helpers/ember_rails_helper.rb
@@ -8,6 +8,11 @@ module EmberRailsHelper
 
     head, body = markup_capturer.capture
 
-    render text: EmberCli[name].index_html(head: head, body: body).html_safe
+    template = EmberCli[name].index_html(head: head, body: body).html_safe
+    begin
+      render text: template
+    rescue ArgumentError
+      render body: template
+    end
   end
 end
--- a/app/helpers/ember_rails_helper.rb
+++ b/app/helpers/ember_rails_helper.rb
@@ -8,11 +8,6 @@ module EmberRailsHelper
 
     head, body = markup_capturer.capture
 
-    template = EmberCli[name].index_html(head: head, body: body).html_safe
-    begin
-      render text: template
-    rescue ArgumentError
-      render body: template
-    end
+    render html: EmberCli[name].index_html(head: head, body: body).html_safe
   end
 end


app/helpers/ember_rails_helper.rb
```diff
@@ -8,6 +8,11 @@ def render_ember_app(name, &block)
 
     head, body = markup_capturer.capture
 
-    render text: EmberCli[name].index_html(head: head, body: body).html_safe
+    template = EmberCli[name].index_html(head: head, body: body).html_safe
+    begin
+      render text: template
+    rescue ArgumentError
+      render body: template
```
seanpdoyle: Could we simplify this block and use `body` as the argument no matter what?

seanpdoyle: It looks like [`render` accepts an `:html` option](https://github.com/rails/rails/blob/e1a72fef7de9a5c0871b0ef1a4acb9483d2bf1f3/guides/source/upgrading_ruby_on_rails.md#rendering-content-from-string), should we use that instead?

weggesser: Sure :html would prob. be better - :body would be the closest equivalent to what :text does/did and I was not sure if the :html option was unwanted in some cases ... For compatibility I chose to have this rescue block - knowing that it's kind of ugly :/ 
4.0 seems not to support :body and :html options for render. http://guides.rubyonrails.org/v4.0/layouts_and_rendering.html#using-render

seanpdoyle: Thanks for clarifying.

This project [supports rails>=4.1](https://github.com/thoughtbot/ember-cli-rails/tree/d03f519ed9c2ae2cb413b6bca76ad04da696fd3d#ruby-and-rails-support), so a breaking change in `4.0` is not a show stopper.

wireframe:
@seanpdoyle any blockers to merging this in?  Rails 5.1 has hit release candidate and it would be awesome to have a ember-cli-rails release w/ compatibility on day one if possible.

seanpdoyle:
@wireframe CI is failing, but it _looks_ like it might be unrelated.

Once it's green, we can merge.

seanpdoyle:
Unfortunately, I've been unable to correct the issues with CI failing.

I've tested this change manually and it's good to go, and has been merged in https://github.com/thoughtbot/ember-cli-rails/commit/585cfea9231232baee544d2402e5c88f12edf0bb.

PR #523, opened 2017-04-19
Cleanup check for html content type

wireframe:
`Mime::Type.html?` is a publicly available API that internally uses a regex.
http://api.rubyonrails.org/classes/Mime/Type.html#method-i-html-3F


seanpdoyle:
Merged in https://github.com/thoughtbot/ember-cli-rails/commit/bd187207177096decc457f103d09ae8bc2b381e6.

Issue #531, opened 2017-06-07
Make `bower install` step optional for ember:install


jakeonfire:
We no longer have any bower dependencies, and yet to use ember-cli-rails we still have a dependency on the bower executable being present (and an empty `bower.json` for that matter). I imagine more ember apps might reach this state as bower itself is kind of being deprecated - the yarn/npm install of bower yields the following post-install message:
```
warning bower@1.8.0: ..psst! While Bower is maintained, we recommend Yarn and Webpack for *new* front-end projects! Yarn's advantage is security and reliability, and Webpack's is support for both CommonJS and AMD projects. Currently there's no migration path but we hope you'll help us figure out one.
```
I imagine doing `if paths.bower` might be a bit risky as in the opposite case (`bower.json` present but no `bower` installed) there would be no warning. But a means of opting out of the `bower prune && bower install` part would be great, and probably only more in-demand as time goes on.

Idea: if there is no `bower.json` do not do `bower install`...?

seanpdoyle:
In light of https://github.com/thoughtbot/ember-cli-rails/pull/532#issuecomment-308141408, I've reopened this issue.

@jakeonfire could you pull down [`0.8.6`](https://github.com/thoughtbot/ember-cli-rails/releases/tag/v0.8.6), try it out, and let me know if it resolves your original issue?

We're going to publish `0.8.7`, which reverts this change, then make another, more robust attempt at resolving the issue, which we'll publish as `0.9.0`.

brandhuge:
Please How can I resolve this in Ubuntu 14.04
# npm install -g bower
npm http GET https://registry.npmjs.org/bower
npm http 304 https://registry.npmjs.org/bower
npm WARN deprecated bower@1.8.0: ..psst! While Bower is maintained, we recommend Yarn and Webpack for *new* front-end projects! Yarn's advantage is security and reliability, and Webpack's is support for both CommonJS and AMD projects. Currently there's no migration path but we hope you'll help us figure out one.
/root/bin/bower -> /root/lib/node_modules/bower/bin/bower
bower@1.8.0 /root/lib/node_modules/bower

seanpdoyle:
@jakeonfire could you try to replicate your issue with the current master build?

```rb
# Gemfile
gem "ember-cli-rails", github: "thoughtbot/ember-cli-rails", branch: "master"
```

jukeboxhero:
[...]

PR #532, opened 2017-06-09
Make `bower` an optional dependency

seanpdoyle:
Closes [#531].

According to `bower` warning messages, the project is semi-deprecated:

```
While Bower is maintained, we recommend Yarn and Webpack for *new*
front-end projects! Yarn's advantage is security and reliability, and
Webpack's is support for both CommonJS and AMD projects. Currently
there's no migration path but we hope you'll help us figure out one.
```

As such, the build process will skip `bower install` and won't require a
`bower` if the application's `bower.json` file is missing.

[#531]: https://github.com/thoughtbot/ember-cli-rails/issues/531

### DIFF
--- a/lib/ember_cli/path_set.rb
+++ b/lib/ember_cli/path_set.rb
@@ -35,6 +35,10 @@ module EmberCli
       @gemfile ||= root.join("Gemfile")
     end
 
+    def bower_json
+      ember_cli_root.join("bower.json")
+    end
+
     def ember
       @ember ||= begin
         root.join("node_modules", "ember-cli", "bin", "ember").tap do |path|
@@ -64,7 +68,7 @@ module EmberCli
     def bower
       @bower ||= begin
         path_for_executable("bower").tap do |bower_path|
-          if bower_path.blank? || !bower_path.executable?
+          if bower_json.exist? && (bower_path.blank? || !bower_path.executable?)
             fail DependencyError.new <<-MSG.strip_heredoc
                 Bower is required by EmberCLI
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 master
 ------
 
+* Don't require `bower` installation if `bower.json` is missing [#532]
+
+[#532]: https://github.com/thoughtbot/ember-cli-rails/pull/532
+
 0.8.5
 -----
 
--- a/README.md
+++ b/README.md
@@ -270,7 +270,8 @@ To configure your EmberCLI-Rails applications for Heroku:
 1. Execute `rails generate ember:heroku`.
 1. Commit the newly generated files.
 1. [Add the NodeJS buildpack][buildpack] and configure NPM to include the
-   `bower` dependency's executable file.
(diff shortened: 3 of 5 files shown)


mike-burns:
Words seem right, code seems right.

lib/ember_cli/path_set.rb
```diff
@@ -63,18 +67,19 @@ def build_error_file
     def bower
       @bower ||= begin
         bower_path = app_options.fetch(:bower_path) { which("bower") }
+        bower_executable = Pathname(bower_path.to_s)
```
geoffharcourt: I was surprised this works! Is there any notable difference between `Pathname.new` and `Pathname()`.

seanpdoyle: I don't remember exactly where I saw this syntax, but it came from [docs](https://ruby-doc.org/stdlib-2.1.2/libdoc/pathname/rdoc/Pathname.html#method-i-each_child).

It's definitely surprising, I'll replace it with `Pathname.new`.

seanpdoyle: @geoffharcourt I plan to merge https://github.com/thoughtbot/ember-cli-rails/pull/533 ahead of this PR.

jdurand:
This just broke my Semaphore CI build with the following even though I have `bower.json` in my ember.js projects.

````
Missing bower packages: 

[...]

Run `bower install` to install missing dependencies.
````

Since this is a breaking change, shouldn't it have been part of a minor version bump?

**Comment by seanpdoyle** [comment]
@jdurand could you try CI again, depending on `gem "ember-cli-rails", github: "thoughtbot/ember-cli-rails", branch: "revert-34ff6dd"` (https://github.com/thoughtbot/ember-cli-rails/pull/535).

If that succeeds, we'll [publish a bugfix version for`0.8.x`](https://github.com/thoughtbot/ember-cli-rails/issues/531#issuecomment-308184038), and bring forward `0.8.6`'s changes to `0.9.0`.

**Comment by jdurand** [comment]
Yes, `revert-34ff6dd` fixes the build.

Note that I was able to fix this by running `bower install` as a postinstall script:

```javascript
//package.json
{
  "scripts": {
    "postinstall": "bower install"
  }
}
```

This could be the 0.9.x way to go as long as it's mentioned in the README and/or CHANGELOG.

jukeboxhero:
Ran into this when trying to remove bower from a project- It looks like `ember:install` returns an exit code 1 when `bower.json` is not present, even though it appears to succeed.

```
> bin/rake RAILS_ENV=staging RAILS_GROUPS=assets ember:install

ember-cli: 2.8.0
node: 6.10.0
os: linux x64
yarn install v1.6.0
[1/5] Validating package.json...
[2/5] Resolving packages...
success Already up-to-date.
Done in 1.41s.

> echo $?
=> 1

```

This is causing capistrano to fail at this step.  If I add a `bower.json` file with blank dependencies to my project, everything works as expected (minus bower being run).  

Not a blocker, but it would be nice to get rid of that `bower.json` file for good 😄

PR #539, opened 2017-06-28
Fix mounting of multiple ember apps in Rails' routes

vfonic:
Because `mount_ember_app :frontend, to: "/"` catches everything, `mount_ember_app :admin_panel, to: "/admin_panel"` would be never reached if specified below the frontend route.


seanpdoyle:
Thanks!

PR #543, opened 2017-07-28
Add support for `rails@5.1`, drop `rails@4.1`

seanpdoyle:
With the release of `rails@5.1`, add explicit test coverage, in addition
to testing on `master`.

According to the [Ruby on Rails Maintenance Policy][policy], Rails 4.1
has reached its end of life stage [after the release of Rails
5.0][rails-5.0].

As such, this commit drops formal test coverage and support for Rails
versions prior to 4.2.

[policy]: http://guides.rubyonrails.org/maintenance_policy.html#severe-security-issues
[rails-5.0]: http://weblog.rubyonrails.org/2016/6/30/Rails-5-0-final/
[rails-5.1]: http://weblog.rubyonrails.org/2017/4/27/Rails-5-1-final/


mike-burns:
LGTM. Is there any leftover code to support older versions of Rails?

seanpdoyle:
@mike-burns some. [Once we drop 4.x we'll be able to remove some more](https://github.com/thoughtbot/ember-cli-rails/blob/f28b903dda7a5ae8c3ceaf37f549eb9d8b49d4ca/lib/ember_cli/deploy/file.rb#L36-L43).

PR #544, opened 2017-07-28
Remove `bin/heroku_install`

seanpdoyle:
For deployments to Heroku, depend on `rake ember:{build,install}` being
executed as part of `rake assets:precompile`.

The removal of `bin/heroku_install` as part of a `postinstall` script in
the root-level `package.json` simplifies support as well.

This commit modifies the setup to incorporate checks for the presence of
`bower.json` to determine whether or not to cache the corresponding
`bower_components` directories, as well as the up-front installation of
the `bower` CLI tool.


nruth:
Getting rid of heroku_install seems like a good change because it's easy to forget to update.
However, this seems to stop heroku from caching the downloaded packages. This doesn't seem worth fixing, but thought I'd point it out.

When the NPM buildpack runs first and caches there's nothing installed yet so nothing to cache, and when the Ruby buildpack (Rails 4.2) runs it only looks for gems and doesn't seem to have customisable cache paths.

I guess it could be addressed by adding custom cache paths to the ruby buildpack, or by adding generic buildpacks that can capture and restore custom cache paths at the start and end of the buildpack stack.

PR #548, opened 2017-09-01
Remove add package version constraint documentation

f1sherman:
Remove documentation suggesting that ember-cli-rails MAJOR and MINOR versions be kept in sync with ember-cli-rails-addon. This documentation appears to be no longer valid since ember-cli-rails-addon has been at 0.8.0 for over a year.

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,16 +97,10 @@ $ cd path/to/frontend
 $ ember install ember-cli-rails-addon
 ```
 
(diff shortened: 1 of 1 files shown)


seanpdoyle:
Thanks for bringing this to our attention. 

Alternatively, would it be better to bump the version on that plugin for each of the major versions that have been published between `0.8.0` and `0.10.0`?

f1sherman:
@seanpdoyle sure, either way works for me! Feel free to close this PR if you like that approach better.

PR #557, opened 2017-12-21
Update README.md with full config/routes.rb path

Dorian:
Makes it clearer / more consistent IMHO.


seanpdoyle:
Thanks for your contribution!

PR #560, opened 2018-03-13
Automatically redirect with a trailing slash when needed

drcapulet:
For apps not mounted at `/`, apps need to write custom redirect login to add the trailing slash Ember requires - this simplified it by providing it as part of `mount_ember_app`.

### DIFF
--- a/lib/ember_cli/route_helpers.rb
+++ b/lib/ember_cli/route_helpers.rb
@@ -15,7 +15,16 @@ module ActionDispatch
         )
 
         scope constraints: ::EmberCli::EmberConstraint.new do
-          get("#{to}(*rest)", routing_options)
+          # Redirect if at the root and missing a trailing slash
+          get(to,
+            to: redirect(Proc.new do |_, request|
+              File.join(request.original_fullpath, "")
+            end),
+            constraints: Proc.new do |request|
+              request.original_fullpath &&
+                !request.original_fullpath.ends_with?("/")
+            end)
+          get(File.join(to, "(*rest)"), routing_options)
         end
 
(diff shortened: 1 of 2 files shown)


seanpdoyle:
Thanks for opening this PR.

Could you rebase off of the latest `master` and try running this change through CI?

drcapulet:
@seanpdoyle looks like that fixed it, thanks!

seanpdoyle:
This change would redirect all Ember routes to corresponding paths ending in `/`.

Could you please elaborate a bit more on your use case? 

Could this the application-global effects break Ember routes that are meant to end without a `/`?

Would `/posts` map to a `posts` route, and `posts/` require a `posts.index` route?

Does Ember routing treat `/posts` equivalent to `posts/`?

drcapulet:
@seanpdoyle this PR only redirects the root URL. If you've got an app mounted at `/admin`, any requests to `/admin` will be redirected to `/admin/` because Ember requires a trailing slash when visiting the root URL - see emberjs/ember.js#4751. Otherwise, the request isn't touched by these changes.

nruth:
I like this. We currently work around this by adding rack-rewrite rules for each mounted app.

seanpdoyle:
Merged in https://github.com/thoughtbot/ember-cli-rails/commit/d5b19df706695d3a0a822f6f0fb5ef37e152840a.

Thanks!

imack:
@seanpdoyle if we're redirecting to a url with query parameters, this seems to have a `/` at the end of it, putting all the query params into the route. ie. `/url?query=test/`

PR #562, opened 2018-03-29
Backtrace is not always 4 spaces long

MadonnaMat:


seanpdoyle:
Thanks for opening this pull request.

Could you add some more information to the commit message?

What sort of output did you encounter to determine this change?

What is your version of `ember-cli`?

Is it possible to test?

MadonnaMat:
I can't seem to replicate it anymore, and I lost the error.txt file, but my system is:
```
ember -v
version: 1.13.13
node: 6.9.4
npm: 2.14.10
os: darwin x64
```
But the issue was that the deprications for one of the ember packages (can't remember which one right now) was outputting with 6 spaces instead of 4.

PR #563, opened 2018-03-30
Replace `poltergeist` with `capybara-selenium`

seanpdoyle:
Improve Appraisal configurations
---

Instead of pinning [`Appraisal`][gem] to a particular patch version of
Rails, pin it to a minor version by declaring versions with the `~>`
prefix.

Additionally, to simplify the generated `.gemfile` files, move
dependencies to the `.gemspec`.

[gem]: https://github.com/thoughtbot/appraisal
 
Upgrade development to `ruby@2.4.4`

The upcoming release of `rails@6.0` (currently tested as `rails@master`)
drops support for Ruby versions older than `2.4`:

```
activesupport-6.0.0.alpha requires ruby version >= 2.4.1,
which is incompatible with the current version, ruby 2.2.2p95
```

While we're upgrading Ruby version, update CI version matrix to use
latest available patch versions.
 
Replace `poltergeist` with `capybara-selenium`
---

 Fix CI Failures by executing feature tests against a headless Chrome
browser instead of PhantomJS.

Configure CI to use the `chrome` stable addon. Add a step to
`bin/setup_ci` to install `chromedriver`.


bin/setup_ci
```diff
@@ -2,6 +2,15 @@
 
 set -e
 
+install_chromedriver() {
+  wget -N http://chromedriver.storage.googleapis.com/2.30/chromedriver_linux64.zip -P ~/
+  unzip ~/chromedriver_linux64.zip -d ~/
+  rm ~/chromedriver_linux64.zip
+  sudo mv -f ~/chromedriver /usr/local/share/
+  sudo chmod +x /usr/local/share/chromedriver
+  sudo ln -s /usr/local/share/chromedriver /usr/local/bin/chromedriver
```
mike-burns: If you pass `-v` to `rm`, `mv`, `chmod`, and `ln`, you can see their output in the CI logs. Might be handy for debugging in the future.

PR #565, opened 2018-04-17
Raise a DependencyError when Yarn is requested but not available

drcapulet:
I ended up hitting some cryptic errors because Yarn was no longer on `$PATH` and thus my dependencies weren't fixed. This change causes any attempted `yarn` commands to fail when `yarn` or `yarn_path` are provided instead of silently falling back to `npm`.

### DIFF
--- a/lib/ember_cli/path_set.rb
+++ b/lib/ember_cli/path_set.rb
@@ -92,7 +92,15 @@ module EmberCli
 
     def yarn
       if yarn?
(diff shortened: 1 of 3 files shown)


seanpdoyle:
Merged as https://github.com/thoughtbot/ember-cli-rails/commit/65151441749adcbbbfb95b1aaf1868a702f384f2.

Thanks!

PR #570, opened 2018-06-26
add install_command_line_arguments to pass through to install commands

wlk125:
By default, yarn and npm will use updated versions of packages when possible.
This behavior may not be desirable in some cases, namely when using build systems
where you want to guarantee a specific version of your package is being used.

This change adds a new `install_command_line_arguments` option to the EmberCLI
initializer that can be used to customize the install command.

### DIFF
--- a/lib/ember_cli/shell.rb
+++ b/lib/ember_cli/shell.rb
@@ -11,6 +11,7 @@ module EmberCli
         options: options,
       )
       @on_exit ||= at_exit { stop }
+      @options = options
     end
 
     def compile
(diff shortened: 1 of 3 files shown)


seanpdoyle:
@wlk125 thanks for looking into this!

Would it make sense to have ` --frozen-lockfile` as part of the default `yarn install` or `npm install` commands?

That would assume that _new_ dependencies or _updated_ versions would be installed elsewhere (like in the Ember sub-directories, in whatever way the appropriate tools dictate.

wlk125:
Definitely for us it would make sense, but not sure if all other users of this gem would agree. It definitely seems like a weird default, even for npm and yarn, to not have their `install` commands work this way.

PR #588, opened 2020-11-08
Ignore query params in trailing slash constraint

straydogstudio:
If someone refreshes a page with query params, a slash is added on the end, breaking the last parameter. This strips anything after a question mark in the url before checking for a trailing slash. 

This appears to only come into play with a root url?

### DIFF
--- a/lib/ember_cli/trailing_slash_constraint.rb
+++ b/lib/ember_cli/trailing_slash_constraint.rb
@@ -1,7 +1,7 @@
 module EmberCli
   class TrailingSlashConstraint
     def matches?(request)
-      !request.original_fullpath.to_s.ends_with?("/")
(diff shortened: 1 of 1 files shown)


lib/ember_cli/trailing_slash_constraint.rb
```diff
@@ -1,7 +1,7 @@
 module EmberCli
   class TrailingSlashConstraint
     def matches?(request)
-      !request.original_fullpath.to_s.ends_with?("/")
+      !request.original_fullpath.to_s.split("?")[0].ends_with?("/")
```
seanpdoyle, line 4: Is there a test case we could add here to both demonstrate the situation and guard against regressions?

straydogstudio:
@seanpdoyle I've been to busy to create a test case. I'll get to it if I can.

PR #592, opened 2021-12-20
Add missing merged PRs to main branch

tricknotes:
It seems some PRs were merged into master branch.
However, now main branch is a default branch.

PR #595, opened 2022-05-30
Redirecting with trailing slash with query parameters

mbackermann:
Fixes the bug when redirecting with a trailing slash with query parameters resulting in an invalid URL
eg: `https://app.com?query=foo` would redirect to `https://app.com?query=foo/`, which is invalid

This PR fixes this issue adding the trailing slash before the query parameters
eg: `https://app.com?query=foo` would be redirected to `https://app.com/?query=foo`

### DIFF
--- a/bin/setup_ember
+++ b/bin/setup_ember
@@ -6,7 +6,7 @@ setup_ember() {
   local target="${1-spec/dummy/my-app}"
 
   if ! [ -d $target ]; then
-    git clone -b stable https://github.com/ember-cli/ember-new-output.git $target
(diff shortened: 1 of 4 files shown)


mbackermann:
by the way, this test is failing on master branch as well

seanpdoyle:
> by the way, this test is failing on master branch as well

The CI execution for the most recent PR merged into `main` (https://github.com/thoughtbot/ember-cli-rails/pull/592) passed. In that time, there haven't been any PRs opened, and no other commits have been merged.

Do you have a sense of what might be at the root of the test failures?

PR #598, opened 2023-02-08
Update CI

tricknotes:
This PR updates the following versions:
- Add Ruby 3.1 & 3.2
- Drop EOL Ruby 2.5 & 2.6
- Update Node.js to 18


seanpdoyle:
Thank you for opening this @tricknotes!

I'm curious if `rails@main` passes as well. Could you also include `main` in the Rails version matrix, and change [this line](https://github.com/thoughtbot/ember-cli-rails/blob/7ce4cefb676fd6cd4c0a8ecd8eb5465998d1d697/Gemfile#L7) from `"master"` to `"main"`?

tricknotes:
Thanks for your review, @seanpdoyle 
I added rails@main into CI matrix. 5fa3c5f592ea9b234d65778dfb24eb740fe532b4

PR #620, opened 2024-08-21
Update default Rails version to the latest of released version

tricknotes:
Using released latest is a good way to develop gem with other gems.


tricknotes:
@seanpdoyle
Could you review this PR?
This is a development-only refactoring.

